PR #7, opened 2023-03-25
chore: Update website on Cargo.toml

jaysonsantos:
Hi there, this is just to update the website because on crates.io it still points to tokio's repository


jtescher:
Sorry for the delay, this should be resolved now.

PR #12, opened 2023-04-01
Update opentelemetry to 0.19

jaysonsantos:
This is a stacked PR with #9.
It should be a clean update, as opentelemetry 0.19 did not break any API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ categories = [
 keywords = ["tracing", "opentelemetry", "jaeger", "zipkin", "async"]
 license = "MIT"
 edition = "2018"
-rust-version = "1.46.0"
+rust-version = "1.56.0"
 
 [features]
 default = ["tracing-log", "metrics"]
@@ -25,7 +25,7 @@ default = ["tracing-log", "metrics"]
 metrics = ["opentelemetry/metrics"]
 
 [dependencies]
-opentelemetry = { version = "0.17.0", default-features = false, features = ["trace"] }
+opentelemetry = { version = "0.18.0", default-features = false, features = ["trace"] }
 tracing = { version = "0.1.35", default-features = false, features = ["std"] }
 tracing-core = "0.1.28"
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["registry", "std"] }
@@ -39,7 +39,7 @@ thiserror = { version = "1.0.31", optional = true }
 [dev-dependencies]
 async-trait = "0.1.56"
 criterion = { version = "0.3.6", default-features = false }
-opentelemetry-jaeger = "0.16.0"
+opentelemetry-jaeger = "0.17.0"
 futures-util = { version = "0.3", default-features = false }
 tokio = { version = "1", features = ["full"] }
 tokio-stream = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tracing-opentelemetry"
-version = "0.19.0"
+version = "0.18.0"
 authors = [
     "Julian Tescher <julian@tescher.me>",
     "Tokio Contributors <team@tokio.rs>"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ opentelemetry-jaeger = "0.18.0"
 futures-util = { version = "0.3", default-features = false }
 tokio = { version = "1", features = ["full"] }
 tokio-stream = "0.1"
+tracing = { version = "0.1.35", default-features = false, features = ["std", "attributes"] }
 
 [lib]
 bench = false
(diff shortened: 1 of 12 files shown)


Oliboy50:
@djc I would say that this is waiting at least for #9

BTW, me too, I would love to see an opentelemetry 0.19 compatible release of tracing-opentelemetry soon... but I understand that moving sources from another repository may have slow things down here... good luck to the contributors 💪

tests/trace_state_propagation.rs
```diff
@@ -158,15 +158,14 @@ fn build_sampled_context() -> (Context, impl Subscriber, TestExporter, TracerPro
...
 impl SpanExporter for TestExporter {
-    async fn export(
-        &mut self,
-        mut batch: Vec<SpanData>,
-    ) -> opentelemetry::sdk::export::trace::ExportResult {
-        if let Ok(mut inner) = self.0.lock() {
-            inner.append(&mut batch);
-        }
-        Ok(())
+    fn export(&mut self, mut batch: Vec<SpanData>) -> BoxFuture<'static, ExportResult> {
```
tomkarw, line 162: This seems like a drive-by change, is there some reason behind it other than personal preference? Does it allow removing async-trait from dependencies?

jaysonsantos: @tomkarw this is from #9 which is an exact backport from the main repo, I am not the original author of it

tomkarw: Got it. It's a nitpick, so I don't think chasing the original author makes sense.

fenollp:
Hi there, looks like this PR needs rebasing! :)

jaysonsantos:
@fenollp this pr is stacked with the 0.18 #9 , I have to wait for that one to be merged first

liamwh:
Perhaps it's also a good idea to include an example with OpenTelemetry OTLP in this PR also, to ensure compilation checks with OpenTelemetry OTLP?

For example, something like the following might serve as a good best practice example base case:

```rs
use opentelemetry_otlp::WithExportConfig;
use tracing_subscriber::{prelude::__tracing_subscriber_SubscriberExt, Layer};

const OTEL_EXPORTER_OTLP_ENDPOINT_ENV_VAR: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
const OTEL_EXPORTER_OTLP_ENDPOINT_DEFAULT: &str = "http://localhost:4317";

const OBSERVABILITY_SERVICE_NAME_ENV_VAR: &str = "OBSERVABILITY_SERVICE_NAME";
const OBSERVABILITY_SERVICE_NAME_DEFAULT: &str = "veloxide-server";

#[tracing::instrument]
pub async fn configure_observability() -> std::result::Result<(), crate::error::Error> {
    let otel_exporter_endpoint =
        dotenvy::var(OTEL_EXPORTER_OTLP_ENDPOINT_ENV_VAR).unwrap_or_else(|_| {
            tracing::warn!(
                "{} Env var not set, using default",
                OTEL_EXPORTER_OTLP_ENDPOINT_DEFAULT
            );
            OTEL_EXPORTER_OTLP_ENDPOINT_DEFAULT.to_string()
        });

    let observability_service_name = dotenvy::var(OBSERVABILITY_SERVICE_NAME_ENV_VAR)
        .unwrap_or_else(|_| OBSERVABILITY_SERVICE_NAME_DEFAULT.to_string());

    let tracer = opentelemetry_otlp::new_pipeline()
        .tracing()
        .with_exporter(
            opentelemetry_otlp::new_exporter()
                .tonic()
                .with_endpoint(otel_exporter_endpoint),
        )
        .with_trace_config(opentelemetry::sdk::trace::config().with_resource(
            opentelemetry::sdk::Resource::new(vec![opentelemetry::KeyValue::new(
                "service.name",
                observability_service_name,
            )]),
        ))
        .install_batch(opentelemetry::runtime::Tokio)?;

    // Create a tracing layer with the configured tracer
    let telemetry_layer = tracing_opentelemetry::layer().with_tracer(tracer);

    let filter = tracing_subscriber::EnvFilter::from_default_env();

    // Use the tracing subscriber `Registry`, or any other subscriber
    // that impls `LookupSpan`
    let subscriber = tracing_subscriber::Registry::default()
        .with(telemetry_layer)
        .with(
            tracing_subscriber::fmt::layer()
                .with_writer(std::io::stdout)
                .with_filter(filter),
        );

    Ok(tracing::subscriber::set_global_default(subscriber)?)
}
```

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tracing-opentelemetry"
-version = "0.17.4"
+version = "0.19.0"
```
jtescher, line 3: Let's update the released version in a separate PR to simplify the merge sequence

```suggestion
version = "0.18.0"
```

PR #26, opened 2023-05-31
Set builder parent span to None when manually setting parent

### LINKED ISSUE #25: `OpenTelemetrySpanExt::set_parent` does not update trace ID in `OtelData`
### Version
```
├── tracing v0.1.37 (*)
├── tracing-opentelemetry v0.18.0
│   ├── tracing v0.1.37 (*)
│   ├── tracing-core v0.1.31 (*)
│   ├── tracing-log v0.1.3
│   │   └── tracing-core v0.1.31 (*)
│   └── tracing-subscriber v0.3.17
│       ├── tracing v0.1.37 (*)
│       ├── tracing-core v0.1.31 (*)
│       ├── tracing-log v0.1.3 (*)
│       └── tracing-serde v0.1.3
│           └── tracing-core v0.1.31 (*)
├── tracing-serde v0.1.3 (*)
└── tracing-subscriber v0.3.17 (*)
```
### Platform

Linux DESKTOP 5.4.72-microsoft-standard-WSL2 #1 SMP Wed Oct 28 23:40:43 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

### Description

I used this code to  create a span and manually change the parent as if I had a remote context to put there:
```
        let span = tracing::trace_span!("Remote trace");
        span.set_parent(opentelemetry::Context::new().with_remote_span_context(
            opentelemetry::trace::SpanContext::new(
                TraceId::from([0x11; 16]),
                SpanId::from([0x22; 8]),
                TraceFlags::SAMPLED,
                true,
                TraceState::default(),
            ),
        ));
```
I have noticed that the `OtelData` associated with the span has different trace IDs in `builder` and `parent_cx`. That is because the span was created with no active span and `OpenTelemetryLayer::on_new_span` created a trace ID. But when a parent context is added, I think this trace ID should be deleted as it should not be used. All other spans with parents also have `None` as their trace ID in the builder.

I do not believe the ID is actually used anywhere, but I think it would be safer to remove it because that would make it harder to misuse this. Unless that would break some guarantee, e.g. that the builder should never change.

mladedav:
Closes #25

## Motivation

When I tried to find a parent span as someone not familiar with the code I tried using what's in the builder and if there's None I tried the parent context (which always has trace ID, it just might be invalid).

This change should simplify this for people unfamiliar with the codebase by getting rid of the trace ID that's not used.

## Solution

Set trace ID in builder to `None`. It has not been used anywhere if parent_cx contained a valid trace ID so no other changes are needed.

### DIFF
--- a/src/span_ext.rs
+++ b/src/span_ext.rs
@@ -124,6 +124,7 @@ impl OpenTelemetrySpanExt for tracing::Span {
                 get_context.with_context(subscriber, id, move |data, _tracer| {
                     if let Some(cx) = cx.take() {
                         data.parent_cx = cx;
+                        data.builder.trace_id = None;
                     }
                 });
             }


jtescher (approved):
Thanks @mladedav

PR #34, opened 2023-07-25
Add set_attribute function

Protryon:
This PR:
* Adds a function `set_attribute` to `OpenTelemetrySpanExt` to allow setting attributes directly on the OpenTelemetry `SpanBuilder`.

This allows users to set more than the tracing maximum of 32 fields.

There was a workaround I came up with without using the function involving wrapping the `Registry` in an Arc and accessing `OtelData` directly, but that led to [this bug in tracing-subscriber](https://github.com/tokio-rs/tracing/issues/2668).


jtescher (approved):
This looks good, thanks @Protryon

eopb:
The 32 field maximum has also been fix over in tracing https://github.com/tokio-rs/tracing/pull/2508

Issue #39, opened 2023-08-02
CI job cargo test (stable on windows-latest) failed to compile pprof


ymgyt:
## Bug Report

CI Job [cargo test (stable on windows-lates)](https://github.com/tokio-rs/tracing-opentelemetry/blob/76771cf8e8a9d13fe9617ed833dbc03e67eff33d/.github/workflows/CI.yml#L111) failed to compile pprof

* [failed example](https://github.com/tokio-rs/tracing-opentelemetry/actions/runs/5732443857/job/15537409319?pr=36)

### Version

`tracing-opentelemetry v0.19.0`

### Platform

Github actions `windows-latest`

### Description

* https://github.com/tikv/pprof-rs/issues/9

PR #40, opened 2023-08-02
chore: use pprof only if target_os is not windows

ymgyt:
## Motivation

Fix https://github.com/tokio-rs/tracing-opentelemetry/issues/39

## Solution

add `#[cfg(not(target_os = "windows"))]` condition to pprof dependencies


jtescher (approved):
Thanks @ymgyt

Issue #32, opened 2023-07-22
MetricsLayer — Add additional event fields as KeyValue pairs to a recorded metric


jnicholls:
## Feature Request

`MetricsLayer` — Add additional event fields as `KeyValue` pairs to a recorded metric.

### Motivation

`MetricsLayer` is a great idea, to have events be able to be captured and exported as metrics. Brilliant. But, the current implementation only submits the metric name and value, and no associated `KeyValue` pairs to add context to the metric.

### Proposal

It would be fantastic if other fields of the event were added as associated `KeyValue` pairs to the metric. Or, another specially named field specifically for `MetricsLayer` to inspect for optional `KeyValue` pairs.

jtescher:
@jnicholls This sounds good! are you able to submit a PR to this effect?

ymgyt:
If @jnicholls  doesn't mind, I'd like to try this feature.

My understanding is that in the [Metric api's Counter add operation spec](https://opentelemetry.io/docs/specs/otel/metrics/api/#add) says

> This API MUST accept the following parameter: 
* A numeric increment value.
* Attributes to associate with the increment value.

This feature allows you to associate the attributes with the metrics.

For example

```rust
tracing::info!
  monotonic_counter.foo = 1,
  key_1 = val_1,
  "foo",
);
```
would export metrics with the attributes of `key_1 = val_1`.

jtescher:
Resolved in #43

PR #43, opened 2023-08-04
feat: produce measurement with attributes

ymgyt:
## Motivation

Support https://github.com/tokio-rs/tracing-opentelemetry/issues/32
This change will allow user to associate Attributes with metrics from tracing.

## Solution

1. MetricsLayer is only interested in events that contain fields with metrics prefixes such as `counter`, so I have utilized the [tracing_subscriber Filter trait](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/#per-layer-filtering).
2. Since it's necessary to pass Attributes when outputting Measurements, it's required to complete the visit for all fields first before invoking the Metrics API(`Counter:add()`).
`SmallVec` is used to hold Attribute KeyValues, Metrics names, and Instruments for invocation, aiming to avoid allocations in typical use cases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ once_cell = "1.13.0"
 # Fix minimal-versions
 async-trait = { version = "0.1.56", optional = true }
 thiserror = { version = "1.0.31", optional = true }
+smallvec = "1.11.0"
 
 [dev-dependencies]
 async-trait = "0.1.56"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ once_cell = "1.13.0"
 # Fix minimal-versions
 async-trait = { version = "0.1.56", optional = true }
 thiserror = { version = "1.0.31", optional = true }
-smallvec = "1.11.0"
+smallvec = "1.0"
 
 [dev-dependencies]
 async-trait = "0.1.56"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ rust-version = "1.60.0"
 [features]
 default = ["tracing-log", "metrics"]
 # Enables support for exporting OpenTelemetry metrics
-metrics = ["opentelemetry/metrics"]
+metrics = ["opentelemetry/metrics", "smallvec"]
 
 [dependencies]
 opentelemetry = { version = "0.20.0", default-features = false, features = ["trace"] }
@@ -35,7 +35,7 @@ once_cell = "1.13.0"
 # Fix minimal-versions
 async-trait = { version = "0.1.56", optional = true }
 thiserror = { version = "1.0.31", optional = true }
-smallvec = "1.0"
+smallvec = { version = "1.0", optional = true }
 
 [dev-dependencies]
 async-trait = "0.1.56"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,10 @@ bench = false
 name = "trace"
 harness = false
 
+[[bench]]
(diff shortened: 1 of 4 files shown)


src/metrics.rs
```diff
@@ -339,19 +368,61 @@ impl MetricsLayer {
...
+                || name.starts_with(METRIC_PREFIX_HISTOGRAM)
+        })
+    }
 }
 
 impl<S> Layer<S> for MetricsLayer
 where
     S: Subscriber + for<'span> LookupSpan<'span>,
 {
     fn on_event(&self, event: &tracing::Event<'_>, _ctx: Context<'_, S>) {
```
ymgyt, line 389: To avoid allocating each time `on_event()` is called, we determine if the event has a metrics field.
However, this caused RwLock::read() to be called for each `on_event()`.

I am concerned that an implementation that gets read lock for every on_event may not be acceptable.

src/metrics.rs
```diff
@@ -339,19 +368,61 @@ impl MetricsLayer {
...
-            instruments: &self.instruments,
-            meter: &self.meter,
-        };
-        event.record(&mut metric_visitor);
+        if self
+            .callsites
+            .has_metrics_field(&event.metadata().callsite())
+        {
+            let mut attributes = Vec::new();
+            let mut visited_metrics = Vec::new();
```
ymgyt, line 395:
```rust
tracing::info!(
  counter.foo = 1,
  counter.bar = 2,
)
```

Since multiple metrics may be output as described above, Vec is used.

I also thought about using thread_local!  to reduce the number of allocations.

src/metrics.rs
jtescher: Yeah we may want to see if there is another way, or have this be an opt-in for non-performance sensitive cases

ymgyt: I will try [per layer filtering](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/#per-layer-filtering)

src/metrics.rs
```diff
@@ -73,64 +83,64 @@ impl Instruments {
...
             }
         };
     }
 }
 
 pub(crate) struct MetricVisitor<'a> {
-    pub(crate) instruments: &'a Instruments,
-    pub(crate) meter: &'a Meter,
+    attributes: &'a mut SmallVec<[KeyValue; 8]>,
+    visited_metrics: &'a mut SmallVec<[(&'static str, InstrumentType); 2]>,
```
ymgyt, line 143: SmallVec size has no specific basis.

ymgyt:
@jtescher 
By  tracing_subscriber Filter trait, I avoid using RwLock. Additionally, I have introduced SmallVec to minimize allocations.

ymgyt:
Here are the benchmark results on my local machine

Screen Shot 2023-08-06 at 16 04 49

Cargo.toml
```diff
@@ -35,6 +35,7 @@ once_cell = "1.13.0"
 # Fix minimal-versions
 async-trait = { version = "0.1.56", optional = true }
 thiserror = { version = "1.0.31", optional = true }
+smallvec = "1.0"
```
jtescher, line 38: nit: can this be option and enabled with the `metrics` feature?

jtescher (approved):
Nice, looks good

ymgyt:
@jtescher Thanks for the review, I made smallvec optional

jtescher:
Looks great, thanks @ymgyt

jnicholls:
Nice work!

Issue #46, opened 2023-08-08
Docs Outdated with V0.20?


stephenandary:
## Bug Report

Following getting started docs imports, I get: 

![image](https://github.com/tokio-rs/tracing-opentelemetry/assets/12686470/9f2ce74b-ba88-43ea-a3a0-738a8c6b2427)

I have yet to dig into changes, and additional docs, but this was not an issue in 0.19.0

Cargo.toml deps related to tracing: 
![image](https://github.com/tokio-rs/tracing-opentelemetry/assets/12686470/2fa064e4-fd79-43d4-a048-3fc94f1f1aa5)

### Version
v0.20.0

stephenandary:
Going to use Dependencies as they are in the cargo.toml here and see what happens first.

stephenandary:
Please include updated documentation indicating the correct dependencies required to use this crate. I have tried to include dependencies in various ways based on what I expect I should need to do, and what is done within this crate. I still hit a lot of issues. 

A cargo.toml showing what to use as a user would go a long way to not hitting so many issues.

jtescher:
Yeah the readme needs to be updated, the lib docs are correct. Do you have time to submit a PR to fix the readme to look more like https://github.com/tokio-rs/tracing-opentelemetry/blob/6bc7dce9560111374dba3ba38e8fdfc257e23541/src/lib.rs#L54C2-L80?

stephenandary:
Not at the moment. I am not currently working on tracing, I just saw the
compile errors today while working on other things as I brought in the
crate for later. I should be able to do so in a week or 2, but not sure
exactly when at the moment.

ymgyt:
I'll try it.

jtescher:
Resolved in #47

PR #47, opened 2023-08-09
docs: update README basic example

ymgyt:
## Motivation

Address https://github.com/tokio-rs/tracing-opentelemetry/issues/46

## Solution

* update example to use opentelemetry-stdout
* add Cargo.toml example

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,14 +59,18 @@ The crate provides the following types:
 ### Basic Usage
 
 ```rust
-use opentelemetry::sdk::export::trace::stdout;
+use opentelemetry::sdk::trace::TracerProvider;
+use opentelemetry::trace::TracerProvider as _;
 use tracing::{error, span};
 use tracing_subscriber::layer::SubscriberExt;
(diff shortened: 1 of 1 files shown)


ymgyt:
@jtescher 
How about adding a complete example of using trace and metrics to crate examples

I believe that tracing-opentelemetry is probably the first repository to be referenced when a tracing user tries to use opentelemetry.

It will be helpful to see a concrete example of using tracing/tracing-subscriber + trucing-opentelemetry + opentelemetry-otlp + opentelemetry in the example of the repository.

If the user understands the specific usage and responsibilities of each crate, they can understand that tracing-opentelemetry only depends on the `TracerProvider` and `MeterProvider` of opentelemetry_api, and the details of usage will be able to refer to opentelemetry-otlp and opentelemetry crate.

On the other hand, this example would duplicate the content of the opentelemetry-otlp and opentelemetry crate examples in many respects.

jtescher (approved):

PR #48, opened 2023-08-10
Move off of opentelemetry in favor of separate crates.

hdost:
## Motivation

`opentelemetry` is going to be deprecated in favor of separated crates.

## Solution

Use the `opentelemetry_sdk` crate for SDK calls instead of `opentelemetry::sdk`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,11 @@ rust-version = "1.60.0"
 [features]
 default = ["tracing-log", "metrics"]
 # Enables support for exporting OpenTelemetry metrics
-metrics = ["opentelemetry/metrics", "smallvec"]
+metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
 
 [dependencies]
 opentelemetry = { version = "0.20.0", default-features = false, features = ["trace"] }
+opentelemetry_sdk = { version = "0.20.0", default-features = false, features = ["trace"] }
 tracing = { version = "0.1.35", default-features = false, features = ["std"] }
 tracing-core = "0.1.28"
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["registry", "std"] }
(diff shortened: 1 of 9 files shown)


shaun-cox (approved):
LGTM

README.md
```diff
@@ -59,16 +59,17 @@ The crate provides the following types:
 ### Basic Usage
 
 ```rust
-use opentelemetry::sdk::trace::TracerProvider;
 use opentelemetry::trace::TracerProvider as _;
+use opentelemetry_sdk::trace::TracerProvider;
+use opentelemetry_sdk::export::trace::stdout;
```
jtescher, line 64: Should revert this, sdk does not have stdout anymore.

Cargo.toml
```diff
@@ -22,10 +22,13 @@ rust-version = "1.60.0"
 [features]
 default = ["tracing-log", "metrics"]
 # Enables support for exporting OpenTelemetry metrics
-metrics = ["opentelemetry/metrics", "smallvec"]
+metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
 
 [dependencies]
 opentelemetry = { version = "0.20.0", default-features = false, features = ["trace"] }
+# TODO: Use blanket implementation in the future over tracer so that
```
jtescher, line 29: Hm I'm not sure that you can implement `PreSampledTracer` in a generic way as it needs to call methods that are not available on the `Tracer` trait. Maybe just remove this TODO for now as the other SDK may have to implement it on their end. We could have the sdk be an optional dependency though potentially to suggest this path.

jtescher (changes requested):

README.md
hdost: Where is it located?

jtescher: it's just a standalone crate like it was before`.with_simple_exporter(opentelemetry_stdout::SpanExporter::default())`

Cargo.toml
hdost: 🆗

README.md
hdost: Yup just realized

jtescher (approved):
Looks good

Issue #54, opened 2023-08-21
Regression with 0.20 regarding remote contexts


valkum:
## Bug Report

We see a regression after updating to v0.20 regarding remote context and how `traceIds` are propagated to children.

### Version

tracing v0.1.37
tracing-futures v0.2.5
tracing-subscriber v0.3.17
tracing-core v0.1.31
tracing-log v0.1.3
tracing-attributes v0.1.26
tracing-serde v0.1.3
tracing-opentelemetry v0.20.0

### Description

We have an instrumented function (`fn handler`) in which we call `OpenTelemetrySpanExt::set_parent`.
Previously, we used something like this:

```rs
let context = Context::default();
let context = TraceContextPropagator::new().extract_with_context(&context, &carrier);
Span::current().set_parent(context);
```

With v0.20 this creates two traces because all spans created after setting the context will get a different `traceId` than the 
span created for `handler`.

The example of `set_parent` uses `TextMapPropagator::extract(&carrier)`. `extract` just calls `Context::currenty()` and then calls `extract_with_context`. If we change our code to use something equivalent:
```rs
let context = Span::current().context();
let context = TraceContextPropagator::new().extract_with_context(&context, &carrier);
Span::current().set_parent(context);
```

we miss a common root span according to Jaeger.

Note that the root span is entered at the time of calling `Span::current().context()`.

I think the regression was caused by https://github.com/tokio-rs/tracing-opentelemetry/pull/26. 

If this is some error at our side, it should be better documented for `set_parent`. If not, and the PR is also not the cause, I assume we hit a regression in opentelemetry.
We are willing to create a PR once we know which path should be taken.

valkum:
I read the description of #45 more thoroughly, and it seems this is a duplicate of #45. Feel free to close this or the other one (depending on which one you think contains more context to fix this)

jtescher:
Resolved in #55

Issue #45, opened 2023-08-08
Correct way to handle trace context propagation when a context might not be present?


shaunbennett:
**edit: perhaps this should have been opened as a Discussion rather than an issue, apologies!**

Greetings, I was hoping for some assistance with handling trace context propagation in cases where a trace context may or may not be present. I've tried digging through the docs and trying a few different options, but ran into issues with each case.

First of all, I should clarify what I expected the result to be:

1. In the case that a trace context exists from `Message`, we utilize that trace context and continue producing spans for the distributed trace
2. In the case where a trace context does not exist from `Message`, we end up generating a new trace with the parent span being `handle_message`

## Attempt 1: Only setting the parent when a trace context is present in `Message`

```rust
#[instrument]
pub fn handle_message(message: Message) {
    if let Some(context) = message.trace_context {
        let context = get_text_map_propagator(|propagator| propagator.extract(context);
        Span::current().set_parent(context);
    }

    // handle the message, generate child spans
}
```

I haven't been able to understand why, but this case ran into a problem where calls between `handle_message` would all bleed into a single trace id. On another thought - I do have a higher level span that's created (at `TRACE` level - with `TRACE` not enabled), could that perhaps be why they end all up being linked together into the same TRACE when I don't propagate a context?

## Attempt 2: Always attempt to extract a context, even when one might not be present

```rust
#[instrument]
pub fn handle_message(message: Message) {
    let context = get_text_map_propagator(|propagator| propagator.extract(message.tracing_context); // may be a value with no trace context fields set
    Span::current().set_parent(context);

    // handle the message, generate child spans
    generate_child_span();
}
```

This resolved the problem of `handle_message` calls without a trace context set all ending up in the same trace, but now I started running into a new issue where the `handle_message` span and `generate_child_span` end up in separate trace ids, but linked to each other. For example, it would look something like this:

```
| ----------------------------------| // handle_message, trace_id: 1, span_id: 1, parent_span_id: None
       |--------------------------|   // generate_child_span, trace_id: 2, span_id: 2, parent_span_id: 1
```

Thus, `generate_child_span` correctly links to `parent_span_id`, but ends up in its own `trace_id` so it doesn't display correctly.

What's the correct way of utilizing `OpenTelemetrySpanExt::set_parent` for these cases?

jtescher:
[...]

PR #55, opened 2023-08-21
fix: trace id did not match when propagating invalid context

ymgyt:
## Motivation

Fix following issues
* https://github.com/tokio-rs/tracing-opentelemetry/issues/54
* https://github.com/tokio-rs/tracing-opentelemetry/issues/45

## Solution

* Revert https://github.com/tokio-rs/tracing-opentelemetry/pull/26/files
* When an invalid(empty) Context is extracted from the propagator and passed to OpenTelemetrySpanExt::set_parent(), if none is set for trace_id, root and child trace id will not match
* I can confirm that this change works in [reproduction](https://github.com/notheotherben/tracing-opentelemetry-traceid-repro) that @notheotherben created
* add regression test case which propagate empty context

### DIFF
--- a/src/span_ext.rs
+++ b/src/span_ext.rs
@@ -143,7 +143,6 @@ impl OpenTelemetrySpanExt for tracing::Span {
                 get_context.with_context(subscriber, id, move |data, _tracer| {
                     if let Some(cx) = cx.take() {
                         data.parent_cx = cx;
-                        data.builder.trace_id = None;
                     }
                 });
             }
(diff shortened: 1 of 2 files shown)


jtescher (approved):
Nice

djc:
@jtescher can we get this published to crates.io? We're also running into this problem at work.

jtescher:
Will be available once https://github.com/tokio-rs/tracing-opentelemetry/pull/56 is approved.

PR #56, opened 2023-08-24
Prepare for v0.21.0 release

jtescher:
Changelog:

### Added

- Ability to produce measurement with attributes (#43)

### Breaking Changes

- `MetricsLayer` is now generic over the its `Subscriber` impl to support [per-layer filtering] (#43)

### Fixed

- Trace IDs not matching when propagating invalid contexts (#55)

[per-layer filtering]: https://docs.rs/tracing-subscriber/0.3.17/tracing_subscriber/layer/index.html#per-layer-filtering

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+# 0.21.0 (August 1, 2023)
+
+### Added
+
+- Ability to produce measurement with attributes (#43)
+
+### Changed
+
+- Depend on `opentelemetry_sdk` directly instead of `opentelemetry::sdk` (#48)
+
+### Fixed
+
+- Trace IDs not matching when propagating invalid contexts (#55)
+
+Thanks to @ymgyt and @hdost for contributing to this release!
+
 # 0.20.0 (August 1, 2023)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,22 @@
-# 0.21.0 (August 1, 2023)
+# 0.21.0 (August 28, 2023)
 
 ### Added
 
 - Ability to produce measurement with attributes (#43)
 
-### Changed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+# 0.21.0 (August 1, 2023)
```
hawkw, line 1: what part of this change is breaking? is the  `opentelemetry_sdk` dependency switch a breaking change, or are the other changes API-breaking? can we document why this is a breaking change version?

djc: Looks like #43 added a type argument to the public `MetricsLayer` type, at least.

jtescher: I don't feel strongly about it, we could consider this a patch release instead for easier upgrades

djc: I suppose the question is, how likely is the extra type argument to break downstream users? But if we don't want to do that analysis, safer to just eat the minor version bump.

hawkw: The added type argument is a breaking change, since it means users who have written e.g. functions that return a `MetricsLayer` will find that their code no longer compiles. I think it would be incorrect to release that change in a patch version.

I'd like it if the changelog could explicitly state what's breaking here, in addition to listing these changes. Can we do that?

CHANGELOG.md
```diff
@@ -1,19 +1,23 @@
-# 0.21.0 (August 1, 2023)
+# 0.21.0 (August 28, 2023)
 
 ### Added
 
 - Ability to produce measurement with attributes (#43)
 
-### Changed
+### Breaking Changes
```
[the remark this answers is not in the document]
jtescher: Yeah seems fine to remove as it's just a re-export

hawkw (approved):

PR #59, opened 2023-09-05
Fix potential deadlock

AsmPrgmC3:
## Motivation

Some `Debug` implementations access a span's extensions, for example a `color_eyre` `Report` when printing a Spantrace.
This can currently lead to a deadlock because `tracing-opentelemetry`'s tracing `Layer` holds an `extensions_mut()` guard for its Span while recording the event.

## Solution

Move the the `OtelData` out of the span to not hold the lock and later write the extension back.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -835,10 +835,9 @@ where
             #[cfg(not(feature = "tracing-log"))]
             let target = target.string(meta.target());
 
-            let mut extensions = span.extensions_mut();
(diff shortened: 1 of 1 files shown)


asonix:
Probably related to: https://github.com/tokio-rs/tracing/issues/1565

jtescher (approved):
Thanks for looking into this, does not appear to have a negative perf impact on existing benchmarks which is a little surprising!

PR #60, opened 2023-09-27
Update README example Cargo.toml

rthomas:
The `Cargo.toml` is missing the `opentelemetry_sdk` dependency to get the `Basic Usage` example working.

## Motivation

Ensure new users to tracing-otel can get up and running easily with the provided example.

## Solution

Update the `Cargo.toml` to the correct set of deps used in the README example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ fn main() {
 ```toml
 [dependencies]
 opentelemetry = "0.20"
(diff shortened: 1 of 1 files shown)


shaun-cox (approved):
Thanks!

jtescher (approved):

PR #78, opened 2023-11-22
Relax opentelemetry dependencies to optimisically use patches

maximebedard:
## Motivation

While trying to submit a bugfix in https://github.com/open-telemetry/opentelemetry-rust, I had versioning issues where the workspace crates would rely on `opentelemetry_sdk = "0.21.1"`, but since it's pinned to "0.21.0" in `tracing-opentelemetry` the resolution would cause conflicts. 

## Solution

I currently relaxed all the `opentelemetry` related crates versions, which should make upgrading a bit easier. Alternatively, I can simply bump the version of `opentelemetry_sdk` to "0.21.1" if that's preferred.

Thank you!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,12 +41,12 @@ smallvec = { version = "1.0", optional = true }
 [dev-dependencies]
 async-trait = "0.1.56"
 criterion = { version = "0.5.1", default-features = false, features = ["html_reports"] }
-opentelemetry = { version = "0.21.0", features = ["trace", "metrics"] }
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

jtescher:
@maximebedard this shouldn't be necessary, the current requirement should be interpreted as `>=0.21.0, <0.22.0` https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

maximebedard:
You are totally right, my bad! Closing.

PR #79, opened 2023-12-13
feat: use `let else` instead of `unwrap`

YangNianYi:


jtescher (approved):
Looks good, thanks @YangNianYi

Issue #77, opened 2023-11-16
Logs do not get attached to `trace_id` when using `#[instrument]


ewoolsey:
Please see [this](https://github.com/open-telemetry/opentelemetry-rust/issues/1378) issue for more context

When looking at the output of the logging pipeline, the `Trace ID` field is blank for all logs emitted inside a function marked by `#[instrument]`

This does not emit a trace id
```rust
#[instrument]
pub async fn my_func()  {
    trace!("fun log");
}
```

But this does
```rust
#[instrument]
pub async fn my_func() {
    let tracer = opentelemetry::global::tracer("ex.com/basic");
    tracer.in_span("cool span", |_| {
        trace!("fun log");
    });
}
```

I'm not sure exactly what's going wrong under the hood.

Here is my initial setup
```rust
pub fn init_tracing() -> Result<()> {
    let stdout = tracing_subscriber::fmt::layer()
        .with_writer(std::io::stdout)
        .compact()
        .without_time()
        .with_target(false)
        .with_line_number(true)
        .with_file(true)
        .with_filter(EnvFilter::from_default_env());

    let subscriber;
    #[cfg(debug_assertions)]
    {
        let console = console_subscriber::spawn();
        subscriber = Registry::default().with(stdout).with(console)
    }
    #[cfg(not(debug_assertions))]
    {
        use crate::metrics::init_meter_provider;
        use opentelemetry::KeyValue;
        use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
        use opentelemetry_sdk::{trace, Resource};
        use tracing_opentelemetry::MetricsLayer;

        let (sender, receiver) = tokio::sync::mpsc::unbounded_channel();
        tokio::spawn(log_handler(receiver));

        let paper_trail = tracing_subscriber::fmt::layer()
            .with_writer(MakePaperTrailWriter { sender })
            .compact()
            .without_time()
            .with_target(false)
            .with_line_number(true)
            .with_file(true)
            .with_filter(EnvFilter::from_default_env());

        let tracing = tracing_opentelemetry::layer()
            .with_tracer(
                opentelemetry_otlp::new_pipeline()
                    .tracing()
                    .with_exporter(opentelemetry_otlp::new_exporter().http())
                    .with_trace_config(trace::config().with_resource(Resource::new(vec![
                        KeyValue::new("service.name", "liquidation-bot-rs"),
                    ])))
[...]

PR #80, opened 2023-12-18
Copy parent trace_id from parent span

maximebedard:
## Motivation

I ran into an issue where the trace_id of the parent span is not copied into the `SpanContext` of the current span. This can be especially problematic when downstream APIs rely on the `OtelData` structure to extract the trace_id and span_id. 

I believe the test was incorrect as it was only checking that the value was left unchanged in the parent span, not that the trace_id of the current span was set.

I believe this is related to https://github.com/tokio-rs/tracing-opentelemetry/issues/77.

## Solution

I'm copying the trace_id from the `parent_cx.span` when the span has a parent context. Otherwise we generate a new trace_id. I've updated the test_tracer to generate sequential `{TraceId,SpanId}` to potentially avoid comparing against `{TraceId,SpanId}::INVALID`.

Thanks!

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -854,10 +854,13 @@ where
             // Eagerly assign span id so children have stable parent id
             .with_span_id(self.tracer.new_span_id());
 
-        // Record new trace id if there is no active parent span
-        if !parent_cx.has_active_span() {
-            builder.trace_id = Some(self.tracer.new_trace_id());
-        }
+        builder.trace_id = if parent_cx.has_active_span() {
+            // Use the parent span trace_id when there's a parent span
+            Some(parent_cx.span().span_context().trace_id())
+        } else {
+            // Otherwise generate a new trace_id
+            Some(self.tracer.new_trace_id())
+        };
 
         let builder_attrs = builder.attributes.get_or_insert(Vec::with_capacity(
             attrs.fields().len() + self.extra_span_attrs(),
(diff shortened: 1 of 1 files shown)


jtescher:
Hm could you elaborate more on the case where parent span ids are not currently copied? Currently the trace id is intentionally only tracked for cases where there is no active parent so it can be used for pre-sampling.

maximebedard:
TIL, I didn't know that's how sampling was handled internally. I haven't looked to much in depth how it's implemented, but I suppose the trace_id is generated when the span is closed instead of when it's created? If that's the case, I'm not really sure how I could solve my problem.

The use-case I currently have is to copy the `trace_id` and `span_id` on log records emitted by using the `opentelemetry-appender-tracing`. I originally had this [PR](https://github.com/open-telemetry/opentelemetry-rust/pull/1394) opened that extracted the trace_id and span_id from the `OtelData` context and would set the `trace_context` based on the span where the log even was emited (See https://github.com/open-telemetry/opentelemetry-rust/pull/1394/files#diff-e9033bd5189bc3b88a31d73e8dfd111a281b76885313fc2c8421b4f9332b452bR153). This allows me to visualize logs and correlate them with their respective spans, which is especially useful when using tools like datadog.

I had to close the original PR due to a circular dependency issue, but I'm using my own version of the `opentelemetry-appender-tracing` that uses published versions of the `tracing-opentelementry` and `opentelemetry-rust` crates.

jtescher:
So the main issue is ordering around calls to `OpenTelemetrySpanExt::set_parent` for spans, e.g. if you were to start a new span, then emit a log record (it would get the current span's trace id `A`) then you call `set_parent` with trace_id `B`, and emit another record, when the span ends it will be exported with trace id `B`, and your first log will be associated with `A` will be confusing. For this reason calls for span contexts go through the `OpenTelemetrySpanExt::context` method which generates a stable trace id if there isn't a parent. If you wanted to get a trace id for your logs you could use the same trick, but you'd have to ensure that there aren't subsequent calls to update the parent context.

maximebedard:
Could a more reliable approach be to buffer the LogRecords, this way we could retrieve a stable trace_id and export them without having to consider calls to `set_parent`?

jtescher:
@maximebedard maybe, you could explore that path and see what tradeoffs you would have to make. Potentially unbounded buffering as logs within a given span aren't capped, also delaying log export until the parent is closed be undesirable. Alternatively we could consider a [special field](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/#special-fields) type solution where if a trace id has been set there it can't be re assigned, and there may be other possible alternative solutions to explore as well.

hdost:
@jtescher the related opentelemetry rust change https://github.com/open-telemetry/opentelemetry-rust/pull/1394

mladedav:
This has been reworked and now changing the parent after some operations should not work anymore (and return an error).

So I'm going to close this but feel free to open an issue if there are still issues with IDs.

PR #85, opened 2024-01-09
Display attribute in metrics

### LINKED ISSUE #81: `Display` attribute not showing up in exported metrics
I'm wondering if this is a bug or a feature. Please allow me to raise this as a bug report in case it's an expected behaviour.

### Version

```sh
$ cargo tree | grep tracing-opentelemetry
├── tracing-opentelemetry v0.21.0
```

I'm using v0.21.0 but the same thing should happen for v0.22.0 too.

### Platform

```sh
uname -a
Darwin Kentas-MacBook-Pro-2.local 22.1.0 Darwin Kernel Version 22.1.0: Sun Oct  9 20:15:52 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T8112 arm64
```

### Description

Currently `MetricsVisitor` does nothing when `record_debug` is called, which happens when an attribute is given as a variable that implements `Display` (and also `Debug`). This results in missing attributes when exported.

I tried this code:

```rust
tracing::info!(
            counter.myapp.organization = 1,
            %status,
);
```

I expected to see this happen: The attribute `status="active"` is exported as a metrics attribute.

Instead, this happened: The attribute doesn't show up.

The reason why this happens is if an attribute is given as something that implements `Display` trait, it's treated as `DisplayValue`, and from the `tracing-core` crate [`Visitor::record_debug` is visited](https://github.com/tokio-rs/tracing/blob/master/tracing-core/src/field.rs#L593) where `MetricsVisitor` doesn't push it to the attributes slice:

https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/metrics.rs#L148-L150

```rust
fn record_debug(&mut self, _field: &Field, _value: &dyn fmt::Debug) {
        // Do nothing
}
```

I'm wondering if this is an expected behaviour or something that needs to be fixed.

Thank you in advance.

KentaKudo:
Closes #81

Adds metrics attributes given as Display or Debug

## Motivation

As described in #81, currently the metrics layer doesn't do anything when Debug / Display values are given like so:

```rust
tracing::info!(
            counter.myapp.organization = 1,
            %organization.status,
            ?organization.type
);
```

This implements `MetricsVisitor::record_debug` so that it can export them as metrics attributes.

## Solution

Push the given argument to the `self.attributes` in `MetricsVisitor::record_debug` where it does nothing currently like it does for str value:

https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/metrics.rs#L148-L150

```rust
fn record_debug(&mut self, _field: &Field, _value: &dyn fmt::Debug) {
        // Do nothing
}
```

https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/metrics.rs#L208-L211

```rust
fn record_str(&mut self, field: &Field, value: &str) {
        self.attributes
            .push(KeyValue::new(field.name(), value.to_owned()));
}
```

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -145,8 +145,9 @@ pub(crate) struct MetricVisitor<'a> {
 }
 
 impl<'a> Visit for MetricVisitor<'a> {
-    fn record_debug(&mut self, _field: &Field, _value: &dyn fmt::Debug) {
-        // Do nothing
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        self.attributes
+            .push(KeyValue::new(field.name(), format!("{value:?}")));
     }
 
     fn record_u64(&mut self, field: &Field, value: u64) {
--- a/tests/metrics_publishing.rs
+++ b/tests/metrics_publishing.rs
@@ -164,22 +164,6 @@ async fn f64_histogram_is_exported() {
     exporter.export().unwrap();
 }
 
-struct DisplayAttribute(String);
-
-impl std::fmt::Display for DisplayAttribute {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "display: {}", self.0)
-    }
-}
-
-struct DebugAttribute(String);
-
-impl std::fmt::Debug for DebugAttribute {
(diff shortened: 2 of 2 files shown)


jtescher (approved):
Looks good thanks @KentaKudo, sorry for the delay.

PR #86, opened 2024-01-09
prevent pancis with per-layer filters

guswynn:
## Motivation

Currently, if the `tracing-opentelemetry` `Layer` has a per-layer filter (in these examples, ones that allow only `>=DEBUG` the following two cases behave differently:

```
let root = tracing::trace_span!("root");
tracing::debug_span!(parent: &root, "child");
```

```
let root = tracing::trace_span!("root");
oot.in_scope(|| tracing::debug_span!("child"));
```

The former panics, and the latter successfully creates a single-span trace. Note that this ONLY happens if there is another layer interested in the `TRACE` level (otherwise, the tracing current-level-filter fast-path filters the root spans out at their callsites).

This may seem rare, but it becomes more common when the per-layer filter is a [reload](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/index.html) filter, which means the parent and child spans can be differently filtered if the filter is reloaded between their creation.

## Solution

Handle this case gracefully. I also did the same in `on_follows_from`


src/layer.rs
```diff
@@ -948,25 +962,27 @@ where
...
-        } else {
-            data.builder.links = Some(vec![follows_link]);
+        // The follows span may be filtered away from this layer, in which case
+        // we just drop the data, as opposed to panicking. This uses the same
+        // reasoning as `parent_context` above.
+        if let Some(follows_span) = ctx.span(follows) {
+            let mut follows_extensions = follows_span.extensions_mut();
+            let follows_data = follows_extensions
+                .get_mut::<OtelData>()
+                .expect("Missing otel data span extensions");
```
guswynn, line 972: Its unclear to me if this also needs to be `if let Some`...my understanding is that if `Context::span` returned span data for the follows span, then this lay did in fact see the follows span and added this data in `on_new_span`. This would mean that the `parent_context` function could also `.expect` getting the `OtelData` from the parent span, but Im not sure...

davidbarsky (approved):
I _think_ this change makes sense, especially since the only behavioral change is "don't panic".

src/layer.rs
jtescher: I am not aware of cases where the contextual spans would not have their data in the registry so I think you are correct here that it's ok.

jtescher (approved):
Looks good, thanks @guswynn, sorry about the delay

guswynn:
@jtescher thanks for the review!

PR #92, opened 2024-02-15
Honor explicit parent spans for events

mladedav:
## Motivation

Currently, when an event is created, no matter what is set as its parent, `OpenTelemetryLayer` will call `lookup_current` and use that as the parent span. It should honor the configured parent, if any, or be ignored, if the parent is explicitly set to `None`.

## Solution

The code is changed so that the parent is correctly resolved. The logic is somewhat similar to `OpenTelemetryLayer::parent_context`.


jtescher (approved):
Makes sense, thanks.

PR #99, opened 2024-02-23
remove i64_histogram support

rbtcollins:
rust-opentelemetry has dropped this, as it is
recommended against by the otel spec.


rbtcollins:
I'm not sure what to do with i64 `histogram.` prefixed metrics - this patch just degrades to a kv store, but this is really a silent API break :/

jtescher (approved):
Want to add a changelog entry for this? Also probably worth emitting a warning or have some behavior or some users may be confused

djc:
This is part of #100.

PR #112, opened 2024-03-03
chore: add contributing guide

mladedav:
## Motivation

The default PR template has a link to a contributing guide that does not exist.

## Solution

I've added a CONTRIBUTING.md file. It's based on `axum`'s with the project name changed as the same (or very similar) template is used across tokio repositories.


mladedav:
This PR is into master because that's where the link in the template leads. Alternatively, we can just change the template.

PR #117, opened 2024-03-17
enhancement: add gauge metric behind feature

academiaresf:
## Motivation

The [opentelemetry](https://github.com/open-telemetry/opentelemetry-rust/blob/a7a47a745dfe9ded79ffb64722fc53847638d47a/examples/metrics-basic/src/main.rs#L112) crate actually supports `Gauge` metric behind `otel_unstable` feature.

## Solution

Extend the current implementation with the treatment for the new `Gauge` metric.

The tests are passing but i don't know if there are necessary other type of tests (at this point, the property `start_time` inside the `DataPoints` of the `Gauge` type are `None` in all the test, for example. L664 of tests/metrics_publishing.rs file).

PR #129, opened 2024-04-30
Metrics/gauge

biryukovmaxim:
Motivation
The [opentelemetry](https://github.com/open-telemetry/opentelemetry-rust/blob/a7a47a745dfe9ded79ffb64722fc53847638d47a/examples/metrics-basic/src/main.rs#L112) crate actually supports Gauge metric behind otel_unstable feature.

Solution
Extend the current implementation with the treatment for the new Gauge metric.

I fixed tests in [pr](https://github.com/tokio-rs/tracing-opentelemetry/pull/117). adds all-features flag to cicd
I also made a pr in the repo of original contributor https://github.com/academiaresf/tracing-opentelemetry/pull/1


src/metrics.rs
```diff
@@ -2,6 +2,8 @@ use std::{collections::HashMap, fmt, sync::RwLock};
 use tracing::{field::Visit, Subscriber};
 use tracing_core::{Field, Interest, Metadata};
 
+#[cfg(feature="otel_unstable")]
```
djc, line 5: Nit: please use spaces around `=` in attributes (for all of the added guards).

src/metrics.rs
```diff
@@ -43,6 +54,12 @@ pub(crate) enum InstrumentType {
     UpDownCounterF64(f64),
     HistogramU64(u64),
     HistogramF64(f64),
+    #[cfg(feature="otel_unstable")]
+    GaugeU64(u64),
+    #[cfg(feature="otel_unstable")]
+    GaugeF64(f64),
+    #[cfg(feature="otel_unstable")]
+    GaugeI64(i64),
```
djc, line 62: Nit: ordering above suggests that `i64` goes before `f64` -- there are no examples of ordering between `u64` and `i64` but it probably makes sense to keep it in that order.

src/metrics.rs
```diff
@@ -125,6 +142,34 @@ impl Instruments {
                     |rec| rec.record(value, attributes),
                 );
             }
+            #[cfg(feature="otel_unstable")]
+            InstrumentType::GaugeF64(value) => {
```
djc, line 146: Nit: order as above.

src/metrics.rs
```diff
@@ -141,13 +186,20 @@ impl<'a> Visit for MetricVisitor<'a> {
...
+            self.visited_metrics
+                .push((metric_name, InstrumentType::GaugeU64(value)));
+            return;
+        }
         if let Some(metric_name) = field.name().strip_prefix(METRIC_PREFIX_MONOTONIC_COUNTER) {
             self.visited_metrics
                 .push((metric_name, InstrumentType::CounterU64(value)));
-        } else if let Some(metric_name) = field.name().strip_prefix(METRIC_PREFIX_COUNTER) {
+            return;
+        }
```
djc, line 199: Not sure why we should drop the `else if` here?

biryukovmaxim: I guess it was done s because there is no way to guard `else if let` by feature gate. I will rollback `else if` as it was, and will add additional gated `if let` at the end of function

src/metrics.rs
Xuanwo: I'm sure rustfmt can handle this. I raised a new issue at https://github.com/tokio-rs/tracing-opentelemetry/issues/143 and is working a PR to fix it.

src/metrics.rs
biryukovmaxim: [fixed](https://github.com/tokio-rs/tracing-opentelemetry/pull/129/commits/d2d0ed2fc3aa88339ce0e722795ca608a92c3102)

src/metrics.rs
biryukovmaxim: [fixed](https://github.com/tokio-rs/tracing-opentelemetry/pull/129/commits/25e7756839d2d61f47e1f736567cb09c4c2ce87e)

src/metrics.rs
biryukovmaxim: [fixed](https://github.com/tokio-rs/tracing-opentelemetry/pull/129/commits/25e7756839d2d61f47e1f736567cb09c4c2ce87e)

src/metrics.rs
biryukovmaxim: [fixed](https://github.com/tokio-rs/tracing-opentelemetry/pull/129/commits/25e7756839d2d61f47e1f736567cb09c4c2ce87e)

Cargo.toml
```diff
@@ -23,6 +23,8 @@ rust-version = "1.65.0"
 default = ["tracing-log", "metrics"]
 # Enables support for exporting OpenTelemetry metrics
 metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
+# Enables support for unstable OpenTelemetry metric instruments
+otel_unstable = ["opentelemetry/otel_unstable"]
```
Xuanwo, line 27: Exporting `otel_unstable` in `tracing-opentelemetry` seems a bit odd to me. 

What do you think about renaming it to something more descriptive like `metrics_gauge_unstable`? This way, we can remove it once `gauge` stabilizes without revealing its origin from `opentelemetry/otel_unstable`.

Xuanwo (commented):
Mostly LGTM

Cargo.toml
biryukovmaxim: yes, sounds reasonable

biryukovmaxim: [here](https://github.com/tokio-rs/tracing-opentelemetry/pull/129/commits/e4cc5f71fcb9857c5d5530d257458b5c2b78ed20)

biryukovmaxim:
one more thing I want to share. Due to [original pr](https://github.com/tokio-rs/tracing-opentelemetry/pull/117) workflow didn't show errors in tests, I added all-features arg to CI, and now it runs tests related to gauge. However it misses the case when it can have compilation errors in case of feature disabled.
something like 
```rust
#[cfg(feature = "feature")]
struct Foo{}

struct Bar(Foo) // not feature gated - compilation error
```
I would suggest using feature matrix in ci to handle it. Can be done [with](https://crates.io/crates/cargo-feature-combinations)

djc (approved):

djc:
Thanks! I think using an external crate tool for this is overkill for now, but it would be nice to add steps that check `--no-default-features` and the default features in addition to the `--all-features` checks. Would be great if you can submit a follow-up PR for this!

PR #144, opened 2024-05-24
chore: Use 2021 edition instead

Xuanwo:
## Motivation

Use 2021 edition instead to make it a bit new.

## Solution

As titled.


Xuanwo:
kindly cc @djc to take a review.

djc:
Nice, thanks!

Issue #146, opened 2024-05-27
Compiling with target `wasm32-wasi` fails


calebschoepp:
### Version

```
├── tracing v0.1.40
│   ├── tracing-attributes v0.1.27 (proc-macro)
│   └── tracing-core v0.1.32
├── tracing-opentelemetry v0.23.0
│   ├── tracing v0.1.40 (*)
│   ├── tracing-core v0.1.32 (*)
│   ├── tracing-log v0.2.0
│   │   └── tracing-core v0.1.32 (*)
│   └── tracing-subscriber v0.3.18
│       ├── tracing-core v0.1.32 (*)
│       └── tracing-log v0.2.0 (*)
└── tracing-subscriber v0.3.18 (*)
```

### Platform

```
Darwin rocinante.local 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64
```

### Description

Compiling an app to a `wasm32-wasi` target fails e.g. `cargo build --target wasm32-wasi --release`. Here's an example of the error I'm seeing:

```
Building component tracing-spin with `cargo build --target wasm32-wasi --release`
   Compiling tracing-opentelemetry v0.23.0
error[E0432]: unresolved import `web_time`
  --> /Users/caleb/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-opentelemetry-0.23.0/src/layer.rs:21:5
   |
21 | use web_time::Instant;
   |     ^^^^^^^^ use of undeclared crate or module `web_time`

error[E0433]: failed to resolve: use of undeclared crate or module `js_sys`
   --> /Users/caleb/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-opentelemetry-0.23.0/src/lib.rs:164:67
    |
164 |         SystemTime::UNIX_EPOCH + std::time::Duration::from_millis(js_sys::Date::now() as u64)
    |                                                                   ^^^^^^ use of undeclared crate or module `js_sys`

Some errors have detailed explanations: E0432, E0433.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `tracing-opentelemetry` (lib) due to 2 previous errors
Error: Build command for component tracing-spin failed with status Exited(101)
```

[...]

PR #142, opened 2024-05-24
Prepare 0.24.0 release

djc:

PR #147, opened 2024-05-28
WASI support

calebschoepp:
## Motivation 
I was attempting to use this library with a WASI target and it was not compiling. The cfg options in the Cargo.toml and the source files did not match. See #146.

## Solution

Fix cfg options to match.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -148,15 +148,18 @@ jobs:
     - name: Run doctests
       run: cargo test --doc --workspace
 
-  test-build-wasm:
-    name: build tests (wasm)
+  test-build-across-targets:
+    name: build tests (${{ matrix.target }})
     needs: check
+    strategy:
+      matrix:
+        target: [wasm32-unknown-unknown, wasm32-wasi]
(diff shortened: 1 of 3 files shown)


djc (approved):
This looks reasonable to me, thanks!

djc:
Can this target be tested in CI? Would be good to avoid regressions.

calebschoepp:
@djc we could perhaps add another job to `CI.yaml` similar to:
```yaml
  test-build-wasm:
    name: build tests (wasm)
    needs: check
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v4
    - uses: dtolnay/rust-toolchain@stable
      with:
        target: wasm32-unknown-unknown
    - name: build all tests
      run: cargo test --no-run
```

I have no familiarity with your CI though so not too sure.

djc:
That looks like a good start, could you add that in?

djc:
Thanks! Going to publish this in #142.

calebschoepp:
@djc thanks for all the reviews and work on this library 👍

PR #148, opened 2024-05-29
chore: fix on_close() comment

Oliboy50:
## Motivation
The PR https://github.com/tokio-rs/tracing-opentelemetry/pull/124 removed the "end time assignment" but the comment was not updated

## Solution
Update the comment


djc (approved):
Thanks!

Issue #150, opened 2024-07-02
Sampling decision cache isn't invalidated on span.set_parent


njsmith:
### Version

```
│   │   ├── tracing-core v0.1.32
│   │   └── tracing-error v0.2.0
│   │       ├── tracing v0.1.40
│   │       │   ├── tracing-attributes v0.1.27 (proc-macro)
│   │       │   └── tracing-core v0.1.32 (*)
│   │       └── tracing-subscriber v0.3.18
│   │           ├── tracing v0.1.40 (*)
│   │           ├── tracing-core v0.1.32 (*)
│   │           └── tracing-log v0.2.0
│   │               └── tracing-core v0.1.32 (*)
│   └── tracing-error v0.2.0 (*)
│   │   │   │   └── tracing v0.1.40 (*)
│   │   │   └── tracing v0.1.40 (*)
│   │   ├── tracing v0.1.40 (*)
│   │   └── tracing v0.1.40 (*)
│   │   │   └── tracing v0.1.40 (*)
│   │   └── tracing v0.1.40 (*)
│       └── tracing v0.1.40 (*)
│   │   │   └── tracing v0.1.40 (*)
│   │   │   └── tracing v0.1.40 (*)
│   └── tracing v0.1.40 (*)
│   │       └── tracing v0.1.40 (*)
│   │   └── tracing v0.1.40 (*)
│   └── tracing v0.1.40 (*)
│   └── tracing-subscriber v0.3.18 (*)
├── tracing v0.1.40 (*)
├── tracing-error v0.2.0 (*)
├── tracing-opentelemetry v0.22.0
│   ├── tracing v0.1.40 (*)
│   ├── tracing-core v0.1.32 (*)
│   ├── tracing-log v0.2.0 (*)
│   └── tracing-subscriber v0.3.18 (*)
├── tracing-subscriber v0.3.18 (*)
│   │   └── tracing v0.1.40 (*)
│   └── tracing v0.1.40 (*)
│   └── tracing v0.1.40 (*)
├── tracing v0.1.40 (*)
├── tracing-error v0.2.0 (*)
├── tracing-opentelemetry v0.22.0 (*)
├── tracing-subscriber v0.3.18 (*)
```

### Platform

Linux njs-0 6.1.58+ #1 SMP PREEMPT_DYNAMIC Mon Jan 29 15:19:25 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Description

[...]

PR #152, opened 2024-07-02
Clear cached sample_result in tracing::Span::set_parent()

edwinsmith:
## Motivation
Fixes https://github.com/tokio-rs/tracing-opentelemetry/issues/150

## Solution
Clear cached `OtelData.builder.sample_result` in `tracing::Span::set_parent()`
so it is recomputed again after the parent changes.

TODO: needs a repro that we can turn into a test.


djc:
Closing in favor of #153, thanks for the effort!

PR #153, opened 2024-07-02
Invalidate sample decision on set parent fix #150

### LINKED ISSUE #150: Sampling decision cache isn't invalidated on span.set_parent

rohitmittapalli:
## Motivation
Fixes https://github.com/tokio-rs/tracing-opentelemetry/issues/150

Test fails without changes and passes with.

### DIFF
--- a/src/span_ext.rs
+++ b/src/span_ext.rs
@@ -143,6 +143,7 @@ impl OpenTelemetrySpanExt for tracing::Span {
                 get_context.with_context(subscriber, id, move |data, _tracer| {
                     if let Some(cx) = cx.take() {
                         data.parent_cx = cx;
+                        data.builder.sampling_result = None;
                     }
                 });
             }
--- a/tests/trace_state_propagation.rs
+++ b/tests/trace_state_propagation.rs
@@ -7,7 +7,7 @@ use opentelemetry::{
 use opentelemetry_sdk::{
     export::trace::{ExportResult, SpanData, SpanExporter},
     propagation::{BaggagePropagator, TraceContextPropagator},
-    trace::{Tracer, TracerProvider},
+    trace::{Config, Sampler, Tracer, TracerProvider},
 };
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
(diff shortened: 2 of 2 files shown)


djc (approved):

djc:
Just needs some formatting fixes in the new test, thanks!

rohitmittapalli:
> Just needs some formatting fixes in the new test, thanks!

Updated and ready for workflow!

djc:
Thanks!

PR #166, opened 2024-09-18
fix: add support for exporting opentelemetry logs_level_enabled feature

HoKim98:
## Motivation

When building `tracing-opentelemetry` and the opentelemetry crate with `logs_level_enabled` feature on a crate, it fails to build on `opentelemetry>=0.25.0`.

It's because the `opentelemetry::logs::Logger` trait contains a fn `event_enabled` which is only enabled with the `logs_level_enabled` feature.

See also:
https://github.com/open-telemetry/opentelemetry-rust/blame/3976f3d4867d564c98a566df83eb59e31c653b84/opentelemetry-sdk/src/logs/log_emitter.rs#L270-L284

## Solution

Add support for exporting opentelemetry logs_level_enabled feature

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,8 @@ rust-version = "1.65.0"
 
 [features]
 default = ["tracing-log", "metrics"]
+# Enables support for exporting OpenTelemetry logs_level_enabled
+logs_level_enabled = ["opentelemetry/logs_level_enabled","opentelemetry_sdk/logs_level_enabled"]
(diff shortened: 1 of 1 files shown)


djc:
So is the motivation here that in your downstream crate you don't depend on opentelemetry-sdk directly at all?

HoKim98:
Ah no, I use both opentelemetry and opentelemetry_sdk crates with `logs_level_enabled` feature.

djc:
So why should this crate be involved in that decision?

PR #169, opened 2024-10-02
Upgrade to opentelemetry 0.26

ymgyt:
## Motivation

Update opentelemetry depencencies to 0.26
https://github.com/open-telemetry/opentelemetry-rust/releases/tag/opentelemetry-0.26.0

## Solution

* Upgrade opentelemetry deps to 0.26
* Adapt to the API changes of OpenTelemetry's `Key`
  * https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/CHANGELOG.md#changelog
* Adapt to the removal of the `AggregationSelector` trait and `DefaultAggregationSelector` struct in `opentelemetry-sdk`.
  * https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md#v0260

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,8 @@ metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
 metrics_gauge_unstable = ["opentelemetry/otel_unstable"]
 
 [dependencies]
-opentelemetry = { version = "0.25", default-features = false, features = ["trace"] }
-opentelemetry_sdk = { version = "0.25.0", default-features = false, features = ["trace"] }
+opentelemetry = { version = "0.26", default-features = false, features = ["trace"] }
+opentelemetry_sdk = { version = "0.26", default-features = false, features = ["trace"] }
 tracing = { version = "0.1.35", default-features = false, features = ["std"] }
 tracing-core = "0.1.28"
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["registry", "std"] }
@@ -41,11 +41,11 @@ smallvec = { version = "1.0", optional = true }
 [dev-dependencies]
 async-trait = "0.1.56"
 criterion = { version = "0.5.1", default-features = false, features = ["html_reports"] }
-opentelemetry = { version = "0.25", features = ["trace", "metrics"] }
-opentelemetry_sdk = { version = "0.25", default-features = false, features = ["trace", "rt-tokio"] }
-opentelemetry-stdout = { version = "0.25", features = ["trace", "metrics"] }
-opentelemetry-otlp = { version = "0.25", features = ["metrics"] }
-opentelemetry-semantic-conventions = "0.25"
+opentelemetry = { version = "0.26", features = ["trace", "metrics"] }
+opentelemetry_sdk = { version = "0.26", default-features = false, features = ["trace", "rt-tokio"] }
+opentelemetry-stdout = { version = "0.26", features = ["trace", "metrics"] }
+opentelemetry-otlp = { version = "0.26", features = ["metrics"] }
+opentelemetry-semantic-conventions = { version = "0.26", features = ["semconv_experimental"] }
 futures-util = { version = "0.3.17", default-features = false }
 tokio = { version = "1", features = ["full"] }
 tokio-stream = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tracing-opentelemetry"
-version = "0.26.0"
+version = "0.27.0"
 description = "OpenTelemetry integration for tracing"
 homepage = "https://github.com/tokio-rs/tracing-opentelemetry"
 repository = "https://github.com/tokio-rs/tracing-opentelemetry"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -38,14 +38,17 @@ lazy_static = { version = "1.0.2", optional = true }
...
-opentelemetry = { version = "0.25", features = ["trace", "metrics"] }
-opentelemetry_sdk = { version = "0.25", default-features = false, features = ["trace", "rt-tokio"] }
-opentelemetry-stdout = { version = "0.25", features = ["trace", "metrics"] }
-opentelemetry-otlp = { version = "0.25", features = ["metrics"] }
-opentelemetry-semantic-conventions = "0.25"
+opentelemetry = { version = "0.26", features = ["trace", "metrics"] }
+opentelemetry_sdk = { version = "0.26", default-features = false, features = ["trace", "rt-tokio"] }
+opentelemetry-stdout = { version = "0.26", features = ["trace", "metrics"] }
+opentelemetry-otlp = { version = "0.26", features = ["metrics"] }
+opentelemetry-semantic-conventions = { version = "0.26", features = ["semconv_experimental"] }
```
ymgyt, line 51: semantic-conventions crate [introduce `semconv_experimental` feature](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-semantic-conventions/CHANGELOG.md#changed)

djc: And why should we enable that here? Do we depend on it in this crate?

Cargo.toml
```diff
@@ -38,14 +38,17 @@ lazy_static = { version = "1.0.2", optional = true }
 thiserror = { version = "1.0.31", optional = true }
 smallvec = { version = "1.0", optional = true }
 
+# https://github.com/open-telemetry/opentelemetry-rust/issues/2165
+tonic = { version = "0.12.3", optional = true }
```
djc, line 42: I think we should just wait for an upstream fix for this instead of working around it like this.

ymgyt, line 42: In OpenTelemetry, [version 0.12 of `tonic` is specified](https://github.com/open-telemetry/opentelemetry-rust/blob/21fc16c0daa3ea8dbede23c814bda4f22d3dfef6/Cargo.toml#L43). However, if version 0.12.2 is used, the compilation fails.

src/layer.rs
```diff
@@ -484,19 +491,31 @@ impl<'a> field::Visit for SpanAttributeVisitor<'a> {
         let error_msg = value.to_string();
 
         if self.sem_conv_config.error_fields_to_exceptions {
-            self.record(Key::new(FIELD_EXCEPTION_MESSAGE).string(error_msg.clone()));
+            self.record(KeyValue::new(
+                Key::new(FIELD_EXCEPTION_MESSAGE),
+                Value::String(error_msg.clone().into()),
```
djc, line 496: Can we do something like `Value::from(error_msg.clone())` instead?

djc (commented):
Thanks!

Cargo.toml
ymgyt: In the [otlp example](https://github.com/tokio-rs/tracing-opentelemetry/blob/ebc0a9209c2c7d3e60212dea7b460becca4915b1/examples/opentelemetry-otlp.rs#L25), a `Resource` was created using semantic conventions. I believe that removing `DEPLOYMENT_ENVIRONMENT_NAME` from the example won't affect its functionality, so if it can be removed, enabling the feature will no longer be necessary.

src/layer.rs
ymgyt: I was able to use `Value::from(error_msg.clone())`

Cargo.toml
djc: Ahh, this is only for the dev-dependencies. I guess that's fine.

Cargo.toml
lalitb: doesn't `cargo update` solve the issue? Seems somehow v0.12.2 is cached locally, and need to be cleaned.

ymgyt: @lalitb Thank you for the suggestion. After using `cargo update` to update `tonic` from 0.12.2 to 0.12.3, the error was resolved.

ymgyt: @djc 
As suggested in the [review](https://github.com/tokio-rs/tracing-opentelemetry/pull/169#discussion_r1785396651) and [issue](https://github.com/open-telemetry/opentelemetry-rust/issues/2165#issuecomment-2389963372), after executing `cargo update`, the error was resolved, and the line added to `Cargo.toml`:

```toml
tonic = { version = "0.12.3", optional = true }
```

became unnecessary.

djc: That's not actually a fix, that's a workaround. Submitted a fix in https://github.com/open-telemetry/opentelemetry-rust/pull/2168.

ymgyt: Thank you for submitting the fix PR. I will update this PR once opentelemetry-proto version 0.26.1 is published.

lalitb: This is released now.

djc (approved):

djc:
Thanks! I'll get this published tomorrow.

djc:
* Published tracing-opentelemetry v0.27.0 at registry `crates-io`
* [new tag]         v0.27.0 -> v0.27.0
* [Release notes](https://github.com/tokio-rs/tracing-opentelemetry/releases/tag/v0.27.0)

Issue #170, opened 2024-10-03
Sync tracing-opentelemetry and opentelemetry crate versions


fabriziosestito:
## Feature Request

It would be nice if  `tracing-opentelemetry`  version was synced with `opentelmetry` version.
For instance, the fact that `tracing-opentelemtry 0.26` depends on `opentelemetry 0.25` and not on `0.26` could lead to confusion.

djc:
This is basically impossible until opentelemetry skips some numbers.

mladedav:
I'm not advocating for this, but what I think `reqwest-tracing` does for example is they have a feature per OpenTelemetry version supported by a given release so when a new version comes out, they can just add a feature and do a patch release. Doing a trick like that once would sync the versions but I don't think it's worth the confusion that the migrations would bring.

One good argument against synchronising is that it would prevent this library to make any breaking changes unless OpenTelemetry also releases breaking changes. Not that this is something that happens often, but it's nice that we have the option if needed.

What I think could help would be re-exporting `opentelemetry` and `opentelemetry_sdk` and discouraging people from depending on them directly. I'm not sure if we should then also add some of the more often used tracers like otlp behind a feature. And people would still need to depend directly if they needed different features from these than we would provide. Do you think that would be viable and could help?

collinpa:
I was bitten by this today, new to the using these crates. I think a note in the readme would have been useful.

I guess there is already this:
```
[dependencies]
opentelemetry = "0.21"
opentelemetry_sdk = "0.21"
opentelemetry-stdout = { version = "0.2.0", features = ["trace"] }
tracing = "0.1"
tracing-opentelemetry = "0.22"
tracing-subscriber = "0.3"
```

but many of the versions didn't match with the current latest so I glossed over it. Updating it to the current latest supported versions might help avoid confusion.

djc:
> I'm not advocating for this, but what I think `reqwest-tracing` does for example is they have a feature per OpenTelemetry version supported by a given release so when a new version comes out, they can just add a feature and do a patch release. Doing a trick like that once would sync the versions but I don't think it's worth the confusion that the migrations would bring.

I feel like this crate has more of a 1:1 mapping from the upstream releases to releases of this crate, so I don't think that approach really makes sense here.

> One good argument against synchronising is that it would prevent this library to make any breaking changes unless OpenTelemetry also releases breaking changes. Not that this is something that happens often, but it's nice that we have the option if needed.

[...]

PR #171, opened 2024-10-08
Added notes to README to clarify version compatibility

jlandahl:
## Motivation

Version compatibility with the OpenTelemetry crates is potentially confusing to new adopters, as the current version of this crate is 0.26, but it is **not** compatible with 0.26 of the OpenTelemetry crates. A naive developer might simply set the versions for all the dependencies to 0.26 in their own project, which results in some very confusing compiler errors that take some research to figure out (ask me how I know!).

## Solution

I added a new section titled "Compatibility with OpenTelemetry crates" just below the "Overview" section. It includes a link to the discussion in #170 as well as a special note about the current 0.26 versus 0.25 situation.

While I was at it, I updated the dependencies in the `Cargo.toml` example in the "Basic Usage" section and ensured that the example code still works with the updated versions.

I also added to the "Visualization example" section, including a link to the `opentelemetry-otlp.rs` example and a minimum `Cargo.toml` example for using that code in a separate project. I simplified the dependencies there to include only the minimum features necessary for the example to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,8 @@ The crate provides the following types:
 
 Note that version numbers for this crate are **not** synchronized with the
 various OpenTelemetry crates, despite having similar version numbers. For
-discussion, see issue #170.
+discussion, see
+[issue #170](https://github.com/tokio-rs/tracing-opentelemetry/issues/170).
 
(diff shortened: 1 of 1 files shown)


djc:
Wanted something slightly different, proposed in #172, I kept you as a co-author on the relevant commits. Thanks!

PR #172, opened 2024-10-09
Update README with versioning notes

djc:
Alternative to #171.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,16 @@ The crate provides the following types:
 
 [msrv]: #supported-rust-versions
 
+## Compatibility with OpenTelemetry crates
+
+Note that version numbers for this crate are **not** synchronized with the
+various OpenTelemetry crates, despite having similar version numbers. For
+discussion, see [issue #170](https://github.com/tokio-rs/tracing-opentelemetry/issues/170).
+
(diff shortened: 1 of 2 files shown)


Xuanwo (approved):
Thank you for polishing the README!

jlandahl:
@djc I see that you moved the basic example out of the readme and into the examples. I think you could have left it in the readme, as it's not taking up a lot of space, and the updates to the `Cargo.toml` snippet was also very useful for new adopters. It also gave a clearl visual example of the versioning differences mentioned in the new note.

I had also added some useful explanatory text around the Jaeger example, including a fully tested `Cargo.toml` snippet that would help anyone who wanted to try that code in their own project. I think that should have stayed, as the full `Cargo.toml` in this repo has a lot of extra dependencies and enabled features which I had pared down to just what was necessary to run that code from an independent project.

Maybe I'll submit a new readme for the `examples` directory with that info.

djc:
@jlandahl code in the README will tend to get out of date because there's no CI checking that it still compiles, so I'd rather have pointers to code that is known to compile. Happy to take further improvements to either the top-level README and/or an extra README for the `examples`!

PR #173, opened 2024-10-12
Reduce locking in the implementation of `Layer::on_close`

mzabaluev:
In the implementation of the `on_close` method, only hold the lock on span's extension to retrieve the extension data, but release it before processing and exporting.

## Motivation

To achieve optimal performance, span locks should be held for as little time as possible.
It's especially troublesome to hold the lock while arbitrary code in the generic OpenTelemetry span and exporter is called to export the span.

## Solution

Reduce the locked code section to what's needed to remove the data of interest from the extension map.


djc (approved):
LGTM, thanks!

PR #176, opened 2024-10-17
adding support for setting the status of the span through SpanExt

webfinesse:
## Motivation

I have found that when working with `tracing` spans created by 3rd party libraries, it is incredibly difficult to set the OpenTelemetry Status of the span, which results in always having an unset span in my visualization tool.

image

In this case, I am working with the span created by the `aws_lambda_runtime` [crate](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-runtime/src/layers/otel.rs#L75).

Unfortunately, since the span in the `aws_lambda_runtime` crate does not include the `otel.status_code` attribute upfront the `OpenTelemetryLayer::SpanAttributeVisitor::record_str` and `OpenTelemetryLayer::SpanAttributeVisitor::record_debug`  never gets triggered when calling `tracing::Span::current().record("otel.status_code", "ok");` This behavior of the `record` function ignoring new fields is documented [here](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.record)

## Solution

My solution to this is add a new member `OpenTelemetrySpanExt::set_status` that takes an `opentelemetry::trace::Status` enum that gets written to the underlying `SpanBuilder`.

### DIFF
--- a/src/span_ext.rs
+++ b/src/span_ext.rs
@@ -1,5 +1,5 @@
 use crate::layer::WithContext;
-use opentelemetry::{trace::SpanContext, Context, Key, KeyValue, Value};
+use opentelemetry::{trace::SpanContext, trace::Status, Context, Key, KeyValue, Value};
 
 /// Utility functions to allow tracing [`Span`]s to accept and return
 /// [OpenTelemetry] [`Context`]s.
@@ -133,6 +133,25 @@ pub trait OpenTelemetrySpanExt {
     /// app_root.set_attribute("http.request.header.x_forwarded_for", "example");
     /// ```
     fn set_attribute(&self, key: impl Into<Key>, value: impl Into<Value>);
+
+    /// Sets an OpenTelemetry status for this span.
+    /// This is useful for setting the status of a span that was created by a library that does not declare
+    /// the otel.status_code field of the span in advance.
+    /// # Examples
+    /// 
+    /// ```rust
+    /// use opentelemetry::trace::Status;
+    /// use tracing_opentelemetry::OpenTelemetrySpanExt;
+    /// use tracing::Span;
+    /// 
+    /// /// // Generate a tracing span as usual
+    /// let app_root = tracing::span!(tracing::Level::INFO, "app_start");
+    ///
+    /// // Set the Status of the span to `Status::Ok`.
+    /// app_root.set_status(Status::Ok);
+    /// 
+    /// ```            
+    fn set_status(&self, status: Status);
 }
 
 impl OpenTelemetrySpanExt for tracing::Span {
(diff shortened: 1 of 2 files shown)


src/span_ext.rs
```diff
@@ -133,6 +133,25 @@ pub trait OpenTelemetrySpanExt {
     /// app_root.set_attribute("http.request.header.x_forwarded_for", "example");
     /// ```
     fn set_attribute(&self, key: impl Into<Key>, value: impl Into<Value>);
+
+    /// Sets an OpenTelemetry status for this span.
```
djc, line 137: Nits: keep the first line in a separate paragraph, which improves formatting in rustdoc. Add an empty line before the `# Examples` header and discard the empty line at the end of the code sample.

src/span_ext.rs
```diff
@@ -207,4 +226,15 @@ impl OpenTelemetrySpanExt for tracing::Span {
             }
         });
     }
+
+    fn set_status(&self, status: Status) {
+        self.with_subscriber(move |(id, subscriber)| {
+            let mut status = Some(status);
```
djc, line 232: Why do you wrap the status in `Option` here?

tests/span_ext.rs
```diff
@@ -0,0 +1,75 @@
...
+        root.set_status(status);
+    });
+
+    drop(provider); // flush all spans
+    let spans = exporter.0.lock().unwrap();
+
+    assert_eq!(spans.len(), 1);
+
+    spans.iter().find(|s| s.name == "root").unwrap().clone()
+}
```
djc, line 75: Nit: add a newline at the end of the file.

djc (commented):
This makes sense to me, thanks!

(In addition to my remarks below, need to fix up the formatting to appease CI.)

src/span_ext.rs
webfinesse: The answer is quite complicated and I was confused myself for a bit. This forum post helped me understand: https://users.rust-lang.org/t/cant-move-a-variable-out-of-fnmut-closure/41497/3

It boils down to the `status` variable gets moved twice. Once in the lambda of `with_subscriber` and again with `with_context`. The `with_context` signature is the following:

```
pub(crate) fn with_context(
        &self,
        dispatch: &tracing::Dispatch,
        id: &span::Id,
        mut f: impl FnMut(&mut OtelData, &dyn PreSampledTracer),
    )
```

`FnMut` is allowed to be called multiple times, and could in theory move `status` multiple times which results in the following error:

```
 fn set_status(&self, status: Status) {
    |                          ------ captured outer variable
...
234 |                 get_context.with_context(subscriber, id, move |builder, _| {
    |                                                          ----------------- captured by this `FnMut` closure
235 |                     builder.builder.status = status;
    |                                              ^^^^^^ move occurs because `status` has type `Status`, which does not implement the `Copy` trait
```

Thus it is telling us that `status` is no longer valid after the first move.

There are few ways to address this.

1 - call `status.clone()`. This will work at the cost of a clone operation because the lambda will have its own independent copy of the status.

2. Put the status into an `Option<T>` and use `take()`. The only reason I did this was there is prior art of this earlier in the file. Notice lines 213-214: 

```
let mut key = Some(key.into());
let mut value = Some(value.into());
```
It is using the `Option<T>` approach to address the same issue, hence to be consistent I used this.

3. See if we can add/change a `with_context` function that accepts a `FnOnce` I tried doing this, but the change kind of spreads through the entire library.

Perhaps there is something I am missing and I am open to other options

tests/span_ext.rs
webfinesse: updated.

src/span_ext.rs
webfinesse: updated.

src/span_ext.rs
djc: That's fair, thanks for the explanation/

Issue #180, opened 2024-10-29
Export level as custom span attribute


mTsBucy1:
### Motivation

I would like to use an opentelemetry backend to collect most of my traces. However, I would also like to be able to glance at them and only see important happenings. `tracing` has the nice concept of levels for this usecase, opentelemetry does not natively, but has custom fields.
This is in relation to the discussion #177.

### Proposal

I propose to also attach the `level` of a `tracing span` to the opentelemetry span.
As far as I can entail, this would been adding an `builder_attrs.push(KeyValue::new("level", attrs.metadata().level()));` to [Layer::on_new_span](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/layer.rs#L846) and possibliy adding a setting to enable this.
(this might be breaking, if it overrides custom "level" attributes).

I'm happy to make this change, but wanted to ask for feedback fist.

### Alternatives

Alternatively, one could always attach an own field, but this entails a lot of duplication.

djc:
I'm open to reviewing a PR for this.

mTsBucy1:
Closed by #182

PR #182, opened 2024-10-29
Add level metadata to spans

mTsBucy1:
See #180 for a description.

This PR implements the `OpenTelemetryLayer::with_level` setting, to disable (enabled by default) attaching the level tag to exported spans.

There are several concerns I see before recommending merging this:
* Will enable by default break code which manually set the `level` field? If so, disable-by-default seems smarter.
* Event levels are still always exported, should we also restrict this?
* I am somewhat confused, but it seems error level *events* set the *span* status to error. Is this behaviour intuitive, once we have a level attached to the span? Should error spans always set the span status?

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -39,6 +39,7 @@ pub struct OpenTelemetryLayer<S, T> {
     location: bool,
     tracked_inactivity: bool,
     with_threads: bool,
+    with_level: bool,
     sem_conv_config: SemConvConfig,
     get_context: WithContext,
     _registry: marker::PhantomData<S>,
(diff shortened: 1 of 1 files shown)


djc:
@mladedav any opinions on this? I don't have a strong opinion.

mladedav:
Looks good to me and I think this will be useful.

This should just also include a changelog entry for anyone who sets some kind of `level` explicitly as a span field.

src/layer.rs
```diff
@@ -562,6 +563,7 @@ where
             location: true,
             tracked_inactivity: true,
             with_threads: true,
+            with_level: true,
```
djc, line 566: I think I'd prefer to leave this as `false` for now.

djc (approved):
@mladedav thanks for the quick followup!

PR #184, opened 2024-11-12
chore: update opentelemetry dependencies to 0.27.0

MathiasPius:
## Motivation

New version 0.27.0 of the opentelemetry crates was recently released.

## Solution

I've updated the dependencies and fixed any outstanding compilation issues caused by breaking changes in the upstream.

Two caveats:

* I was not able to resolve the compilation issues with https://github.com/tokio-rs/tracing-opentelemetry/blob/95f9ab97e239e5789d71b9636445d89e47612f65/benches/metrics.rs since the `NoopMeterProvider` has been hidden: https://github.com/open-telemetry/opentelemetry-rust/pull/2191
* The metric "views" are now gated behind the feature `spec_unstable_metrics_views`, and I wasn't sure whether to still include it in the example: https://github.com/MathiasPius/tracing-opentelemetry/blob/d18214596c97b1243ee955a63bed3875a4f8f5f9/examples/opentelemetry-otlp.rs#L47-L82

### DIFF
--- a/examples/opentelemetry-otlp.rs
+++ b/examples/opentelemetry-otlp.rs
@@ -1,11 +1,8 @@
-use opentelemetry::{global, trace::TracerProvider, Key, KeyValue};
+use opentelemetry::{global, trace::TracerProvider as _, KeyValue};
 use opentelemetry_sdk::{
-    metrics::{
-        reader::DefaultTemporalitySelector, Aggregation, Instrument, MeterProviderBuilder,
-        PeriodicReader, SdkMeterProvider, Stream,
-    },
+    metrics::{MeterProviderBuilder, PeriodicReader, SdkMeterProvider},
     runtime,
-    trace::{BatchConfig, RandomIdGenerator, Sampler, Tracer},
+    trace::{RandomIdGenerator, Sampler, Tracer, TracerProvider},
     Resource,
 };
 use opentelemetry_semantic_conventions::{
@@ -30,9 +27,10 @@ fn resource() -> Resource {
 
 // Construct MeterProvider for MetricsLayer
 fn init_meter_provider() -> SdkMeterProvider {
-    let exporter = opentelemetry_otlp::new_exporter()
-        .tonic()
-        .build_metrics_exporter(Box::new(DefaultTemporalitySelector::new()))
+    let exporter = opentelemetry_otlp::MetricExporter::builder()
+        .with_tonic()
+        .with_temporality(opentelemetry_sdk::metrics::Temporality::default())
+        .build()
         .unwrap();
 
     let reader = PeriodicReader::builder(exporter, runtime::Tokio)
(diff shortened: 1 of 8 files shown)


douggynix:
In order to solve the compilation issue from **_benches/metrics.rs_** with **NoopMeterProvider**,
you can declare NoopMeterProvider for the purpose of that metrics. it doesn't hurt replicating the logic in here. 
I did it like this, it compiles successfully and the test was being to run successfully. Hope that helps for the compilation errors.
```rust
use opentelemetry::metrics::{InstrumentProvider, Meter, MeterProvider};

struct NoopMeterProvider{
    _private: ()
}

impl NoopMeterProvider {
    fn new()->Self{
        Self{
            _private: ()
        }
    }
}

impl MeterProvider for NoopMeterProvider{
    fn meter_with_scope(&self, scope: InstrumentationScope) -> Meter {
        Meter::new(Arc::new(NoopMeter::new()))
    }
}

struct NoopMeter {
    _private: (),
}

impl NoopMeter {
    /// Create a new no-op meter core.
    pub(crate) fn new() -> Self {
        NoopMeter { _private: () }
    }
}

impl InstrumentProvider for NoopMeter {}

fn metrics_events(c: &mut Criterion) {
.........
```

MathiasPius:
That was also the solution I was considering.

With a maintainer's blessing and some guidance on how to resolve the `spec_unstable_metrics_views` change, I can make the changes later today :)

MathiasPius:
I've resolved the caveats in the OP:

* Added implementation of `NoopMeterProvider` in `benches/metrics.rs`
* Enabled `spec_unstable_metrics_views` feature for `opentelemetry-sdk` (in *dev-dependencies* only)

examples/opentelemetry-otlp.rs
```diff
@@ -41,11 +41,13 @@ fn init_meter_provider() -> SdkMeterProvider {
 
     // For debugging in development
     let stdout_reader = PeriodicReader::builder(
-        opentelemetry_stdout::MetricsExporter::default(),
+        opentelemetry_stdout::MetricExporter::default(),
         runtime::Tokio,
     )
     .build();
 
+    // NOTE: Views are gated behind opentelemetry's `spec_unstable_metrics_views` feature.
```
djc, line 49: Thanks for digging into this, but while views are unstable I think we should avoid showing them off in our examples.

djc (commented):
Thanks for doing the legwork on this!

Would you mind adding a separate commit with a version bump?

examples/opentelemetry-otlp.rs
MathiasPius: Done!

MathiasPius:
* Removed the "views" functionality from the example
* Fixed the cargo hack powerset feature failures, missed some `.init()` to `.build()` renames.
* Bumped version to `0.28.0`

djc:
Looks like you'll need to bump the MSRV:

https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/Cargo.toml#L16
https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/.github/workflows/CI.yml#L91

djc:
* Published tracing-opentelemetry v0.28.0 at registry `crates-io`
* [new tag]         v0.28.0 -> v0.28.0
* [Release notes](https://github.com/tokio-rs/tracing-opentelemetry/releases/tag/v0.28.0)

douggynix:
Thanks

PR #188, opened 2025-01-28
use the tracing event name as the otel event name

zzlk:
## Motivation

When using the tracing macros and trying to explicitly set the 'name' of the event, for example:
```rust
event!(name: "test", Level::INFO, field1 = "field1");
```
This event will get emitted to open telemetry as an event with an empty name. The 'name' information will be completely lost. It seems common for this information to get dropped actually, see: https://github.com/tokio-rs/tracing/issues/2774

## Solution

In on_event I set the event name to `event.metadata().name()` which should be the name of the event that is handed down from tracing. I am unfamiliar with this code and the wider tracing/opentelemtry ecosystem so I am not sure this is a good way to solve this issue, or if this is even a real issue. I am also not sure what this change would break, and what kind of backward compatibility guarantees there might be. I have tested my use case (sending traces/events to jaeger) and can confirm that the names of events are being recorded correctly with this change.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -1071,6 +1071,15 @@ where
                 sem_conv_config: self.sem_conv_config,
             });
 
+            // If the event name is still empty, then there was no special handling of error fields.
+            // It should be safe to set the event name to the name provided by tracing.
+            // This is a hack but there are existing hacks that depend on the name being empty, so to avoid breaking those the event name is set here.
+            // Ideally, the name should be set above when the event is constructed.
+            // see: https://github.com/tokio-rs/tracing-opentelemetry/pull/28
+            if otel_event.name.is_empty() {
+                otel_event.name = std::borrow::Cow::Borrowed(event.metadata().name());
+            }
+
             let mut extensions = span.extensions_mut();
             let otel_data = extensions.get_mut::<OtelData>();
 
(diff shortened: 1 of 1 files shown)


src/layer.rs
```diff
@@ -1052,6 +1052,7 @@ where
             let target = KeyValue::new(target, Value::String(meta.target().into()));
 
             let mut otel_event = otel::Event::new(
+                // event.metadata().name(),
```
djc, line 1055: This comment looks like you've commented out an additional argument, which is really confusing. Please move it to the right of the `String::new()` argument.

src/layer.rs
```diff
@@ -1489,6 +1499,45 @@ mod tests {
...
+            .with_parent("user error");
+
+        tracing::subscriber::with_default(subscriber, || {
+            let span = tracing::debug_span!(
+                "request",
+                error = &err as &(dyn std::error::Error + 'static)
+            );
+
+            span.in_scope(|| {
+                tracing::event!(tracing::Level::INFO, "event name 1");
```
djc, line 1518: I'm surprised that `"event name 1"` ends up in `metadata.name()`. Why is that? It would seem like this ought to be the `message` field.

djc (commented):
I think this might make sense.

src/layer.rs
zzlk: The message field is copied to the event name on line 219 and not added as an otel attribute, so message = "x" is essentially equivalent to name: "x".

src/layer.rs
zzlk: This was left in accidentally. I have removed it since there is already a comment in the code below explaining the ideal possibility of setting the name at this location.

djc (approved):

PR #193, opened 2025-02-11
Update opentelemetry to 0.28

cschramm:


djc:
Thanks for working on this!

Can you add a commit that bumps the MSRV for this project to 1.75 in both CI and the Cargo metadata?

djc:
Sorry, and also, a commit that bumps the version number?

cschramm:
Sure, you're welcome. :smile:

BSteffaniak:
Wow, just did the same. lol. I'll close mine too.

BSteffaniak:
Did you run into the error that I ran into while doing my PR?

> There is an issue with the SdkMeterProvider not properly shutting down. The root of the error I was able to see said this: Status { code: Unimplemented, message: "unknown service opentelemetry.proto.collector.metrics.v1.MetricsService", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }. I am not super knowledgeable around this area, so I'm not sure what exactly needs to be fixed here, BUT, it seems to also happen with the [basic-oltp](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/examples/basic-otlp/src/main.rs) example itself, so it seems like the error is likely out of scope of this repo. I'm going to open an issue there for that error and maybe we can get that figured out too.

Cargo.toml
```diff
@@ -13,18 +13,18 @@ categories = [
...
 edition = "2021"
-rust-version = "1.70.0"
+rust-version = "1.75.0"
 
 [features]
 default = ["tracing-log", "metrics"]
 # Enables support for exporting OpenTelemetry metrics
 metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
 # Enables experimental support for OpenTelemetry gauge metrics
-metrics_gauge_unstable = ["opentelemetry/otel_unstable"]
```
cschramm, line 23: Note: This feature was necessary with opentelemetry 0.22, still in use when #129 was merged, but already obsolete with opentelemetry 0.23, so that it was not actually necessary in any tracing-opentelemetry release.

djc:
Thanks!

* Published tracing-opentelemetry v0.29.0 at registry `crates-io`
* [new tag]         v0.29.0 -> v0.29.0
* [Release notes](https://github.com/tokio-rs/tracing-opentelemetry/releases/tag/v0.29.0)

PR #195, opened 2025-03-12
doc: tracked inactivity default

aimericsr:
## Motivation

The documentation doesn't provide the default value of tracked_inactivity for `OpenTelemetryLayer` .

## Solution

Specify that, by default, inactivity tracking is enabled.


djc (approved):

djc:
Thanks!

PR #196, opened 2025-03-22
chore: update from opentelemetry-sdk 0.28.0 to 0.29.0

alessandrobologna:
## Motivation

This PR updates the dependencies from opentelemetry 0.28.0 to 0.29.0. The update requires adapting to breaking changes in the `SpanExporter` trait interface, which now requires immutable references (`&self` instead of `&mut self`) and uses `async fn` syntax rather than returning `BoxFuture`.

## Solution

- Updated opentelemetry dependencies from 0.28.0 to 0.29.0 in Cargo.toml
- Updated all `SpanExporter` trait implementations in test files and examples to match the breaking changes in the 0.29.0 release:
  - Changed `&mut self` to `&self` in method signatures
  - Changed return type from `BoxFuture<'static, OTelSdkResult>` to using `async fn` that returns `OTelSdkResult` directly
  - Removed unnecessary `BoxFuture` imports
- Updated CHANGELOG.md with the breaking change

The specific breaking changes in opentelemetry-sdk 0.29.0 that affected this crate are documented in the [upstream changelog](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md#0290):

> **Breaking** The SpanExporter::export() method no longer requires a mutable reference to self. Before:  
> `async fn export(&mut self, batch: Vec<SpanData>) -> OTelSdkResult`  
> After:  
> `async fn export(&self, batch: Vec<SpanData>) -> OTelSdkResult`

All tests and examples have been verified to build and run correctly with these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+### Breaking Changes
+
+- Upgrade from opentelemetry 0.28.0 to 0.29.0. Refer to the upstream
+  [changelog](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md#0290)
+  for more information.
+
 # 0.27.0 (October 9, 2024)
 
 ### Breaking Changes
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,8 @@ metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
 metrics_gauge_unstable = []
 
 [dependencies]
-opentelemetry = { version = "0.28.0", default-features = false, features = ["trace"] }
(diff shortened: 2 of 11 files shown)


djc (approved):

djc:
Thanks! Looks like we this might also need an MSRV adjustment...

djc:
> Thanks! Looks like we this might also need an MSRV adjustment...

Never mind, the MSRV was already high enough. Just fix the formatting!

alessandrobologna:
Sorry, I missed that, pushed another commit with the formatting fix. Thank you.

tisonkun:
Can we have a release for this patch?

djc:
Sure - can you submit a PR with a version bump?

PR #201, opened 2025-04-26
feat: Add OpenTelemetrySpanExt methods for direct span event creation

### LINKED ISSUE #200: Feature Request: Support for Adding OpenTelemetry Span Events with Dynamic Attributes
### Motivation

Currently, the primary way to add span events within `tracing-opentelemetry` is via the `tracing::event!` macro. While this works well for statically defined attributes, it presents a limitation when needing to record span events where attribute keys are determined at runtime.

**Use Case:** Processing arbitrary structured data (e.g., JSON payloads for configuration updates, user interactions) where the specific fields within the payload are not known at compile time. The goal is to record these dynamic key-value pairs as distinct OpenTelemetry span event attributes so that they can be used for querying and filtering in the various observability backends.

**Problem:** The `tracing::event!` macro requires static field names. The only workaround for dynamic keys is to serialize the attributes (e.g., to JSON) into a single string field. This is often not optimal, as it may prevent or make it more difficult to do filtering, aggregation, or searching based on the actual dynamic keys within the backend system. It also makes it more difficult to port otel instrumented code from other languages to rust, or creating internal libraries that provide similar features across platforms.

### Proposal

I propose extending the `OpenTelemetrySpanExt` trait with methods to directly add OpenTelemetry span events, mirroring the flexibility of the underlying OpenTelemetry API:

```rust
trait OpenTelemetrySpanExt {
    // ... existing methods ...

    fn add_otel_span_event(&self, name: String, attributes: Vec<KeyValue>);
    fn add_otel_span_event_with_timestamp(&self, name: String, timestamp: SystemTime, attributes: Vec<KeyValue>);
}
```
The trait methods could be just called `add_event` for consistency with the otel Span trait, but maybe that could be confusing with tracing event, so I thought a more specific name may make more sense in this context.

**Implementation:** These methods would leverage the existing `with_subscriber` -> `downcast_ref::<WithContext>` -> `with_context` pattern (similar to `set_attribute` and `set_status`) to access the internal `OtelData` for the span and append a constructed `opentelemetry::trace::Event` to its `builder.events` list.

**Why:** This provides a direct, idiomatic way within the `tracing-opentelemetry` ecosystem to handle span events with dynamic attributes, avoiding lossy serialization. It aligns with the precedent set by `set_attribute` for handling dynamic span data.

**Drawbacks:** Minimal drawbacks anticipated. It adds two methods to the trait surface area. It relies on the `with_subscriber` mechanism working correctly in the user's setup.

### Alternatives

1.  **`tracing::event!` + Serialization:**
    *   *Drawback:* As mentioned, serializing dynamic attributes into a string significantly harms observability and querying capabilities.
2.  **Manual OpenTelemetry Context Management:**
    *   *Drawback:* Managing OpenTelemetry context manually alongside `tracing` context is complex, error-prone, and would certainly defeat much of the purpose of using `tracing-opentelemetry`.

I think the addition to `OpenTelemetrySpanExt` is preferred because it directly addresses the limitation for dynamic attributes within the existing crate structure, is consistent with other extension methods like `set_attribute`, or `add_link` and avoids the major drawbacks of the alternatives.

*(I have a working implementation of this in a local fork and would be happy to open a PR if this feature request is accepted.)*

alessandrobologna:
This PR adds methods to `OpenTelemetrySpanExt` for directly adding OpenTelemetry span events, addressing the limitations of using `tracing::event!` for events with dynamic attributes.

Fixes #200

## Motivation
As discussed in issue #200, the current `tracing::event!` macro requires attribute keys to be known at compile time. This makes it difficult or impossible to record OpenTelemetry span events with dynamic attributes (e.g., from processing arbitrary JSON payloads) without resorting to lossy serialization, which hinders observability.

This change provides a way to directly add `opentelemetry::trace::Event` objects with dynamic `Vec<KeyValue>` attributes, similar to how `set_attribute` allows setting dynamic span attributes. This improves flexibility, makes porting code easier, and aligns better with the capabilities of the OpenTelemetry Span API.

## Solution
This PR extends the `OpenTelemetrySpanExt` trait with two new methods:

1.  `add_otel_span_event<T>(&self, name: T, attributes: Vec<KeyValue>) where T: Into<Cow<'static, str>>`: Adds an OpenTelemetry span event with the current timestamp.
2.  `add_otel_span_event_with_timestamp<T>(&self, name: T, timestamp: SystemTime, attributes: Vec<KeyValue>) where T: Into<Cow<'static, str>>`: Adds an OpenTelemetry span event with a specified timestamp.

The implementation uses the existing `with_subscriber` -> `downcast_ref::<WithContext>` -> `with_context` pattern found in other `OpenTelemetrySpanExt` methods (`set_attribute`, `set_status`) to access the internal `OtelData` builder and push the event.

The name `add_otel_span_event` was chosen over `add_event` to avoid potential confusion with the `tracing::event!` macro within the context of this crate.

This PR includes:
- The trait method definitions and implementations in `src/span_ext.rs`.
- Corresponding test cases in `tests/span_ext.rs`.
- Updates to `README.md` documentation.
- A new example file `examples/opentelemetry-span-event.rs`.
- Updated `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+### Added
+
+- Add `OpenTelemetrySpanExt::add_event` and `OpenTelemetrySpanExt::add_event_with_timestamp` 
+  functions to allow adding OpenTelemetry events directly to a `tracing::Span`, enabling the use of dynamic attribute keys 
+  and custom event timestamps.
+
 # 0.30.0 (March 23, 2025)
 
 ### Breaking Changes
--- a/src/span_ext.rs
+++ b/src/span_ext.rs
@@ -1,5 +1,10 @@
 use crate::layer::WithContext;
-use opentelemetry::{trace::SpanContext, trace::Status, Context, Key, KeyValue, Value};
+use opentelemetry::{
+    time,
+    trace::{SpanContext, Status},
+    Context, Key, KeyValue, Value,
+};
+use std::{borrow::Cow, time::SystemTime};
 
 /// Utility functions to allow tracing [`Span`]s to accept and return
 /// [OpenTelemetry] [`Context`]s.
@@ -152,20 +157,76 @@ pub trait OpenTelemetrySpanExt {
     /// app_root.set_status(Status::Ok);
     /// ```            
     fn set_status(&self, status: Status);
+
+    /// Adds an OpenTelemetry event directly to this span, bypassing `tracing::event!`.
+    /// This allows for adding events with dynamic attribute keys, similar to `set_attribute` for span attributes.
+    /// Events are added with the current timestamp.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use opentelemetry::{KeyValue};
+    /// use tracing_opentelemetry::OpenTelemetrySpanExt;
+    /// use tracing::Span;
+    ///
+    /// let app_root = tracing::span!(tracing::Level::INFO, "processing_request");
+    ///
+    /// let dynamic_attrs = vec![
+    ///     KeyValue::new("job_id", "job-123"),
+    ///     KeyValue::new("user.id", "user-xyz"),
+    /// ];
+    ///
+    /// // Add event using the extension method
+    /// app_root.add_event("job_started".to_string(), dynamic_attrs);
+    ///
+    /// // ... perform work ...
+    ///
+    /// app_root.add_event("job_completed", vec![KeyValue::new("status", "success")]);
+    /// ```
+    fn add_event(&self, name: impl Into<Cow<'static, str>>, attributes: Vec<KeyValue>);
+
+    /// Adds an OpenTelemetry event with a specific timestamp directly to this span.
+    /// Similar to `add_event`, but allows overriding the event timestamp.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use opentelemetry::{KeyValue};
+    /// use tracing_opentelemetry::OpenTelemetrySpanExt;
+    /// use tracing::Span;
+    /// use std::time::{Duration, SystemTime};
+    /// use std::borrow::Cow;
+    ///
+    /// let app_root = tracing::span!(tracing::Level::INFO, "historical_event_processing");
+    ///
+    /// let event_time = SystemTime::now() - Duration::from_secs(60);
+    /// let event_attrs = vec![KeyValue::new("record_id", "rec-456")];
+    /// let event_name: Cow<'static, str> = "event_from_past".into();
+    ///
+    /// app_root.add_event_with_timestamp(event_name, event_time, event_attrs);
+    /// ```
+    fn add_event_with_timestamp(
+        &self,
+        name: impl Into<Cow<'static, str>>,
+        timestamp: SystemTime,
+        attributes: Vec<KeyValue>,
+    );
 }
 
 impl OpenTelemetrySpanExt for tracing::Span {
     fn set_parent(&self, cx: Context) {
         let mut cx = Some(cx);
         self.with_subscriber(move |(id, subscriber)| {
-            if let Some(get_context) = subscriber.downcast_ref::<WithContext>() {
-                get_context.with_context(subscriber, id, move |data, _tracer| {
-                    if let Some(cx) = cx.take() {
-                        data.parent_cx = cx;
-                        data.builder.sampling_result = None;
-                    }
-                });
-            }
+            let Some(get_context) = subscriber.downcast_ref::<WithContext>() else {
+                return;
+            };
+            get_context.with_context(subscriber, id, move |data, _tracer| {
+                let Some(cx) = cx.take() else {
+                    return;
+                };
+                data.parent_cx = cx;
+                data.builder.sampling_result = None;
+            });
         });
     }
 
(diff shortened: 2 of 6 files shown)


alessandrobologna:
Hi @djc (and other maintainers),
Just wanted to add a couple of notes about this PR:
Clippy Fix Commit: You'll notice the last commit (c54ba96) addresses a Clippy lint (`double_ended_iterator_last` in `tests/metrics_publishing.rs`). This warning was surfaced by the CI checks (-D warnings) and appeared unrelated to the core feature addition, so I kept it as a separate commit for clarity.
Versioning & Changelog: I haven't bumped the crate version in Cargo.toml, and the changelog entries for the new feature and example are currently under the ## Unreleased section in CHANGELOG.md.

I'm happy to handle the version bump and finalize the changelog entry if after your review this PR is ready to merge. I wasn't sure about the appropriate version target for this addition, since it's adding new API surface (non-breaking) to a 0.x crate, would you typically target a minor version bump (e.g., 0.31.0) or a patch release (e.g., 0.30.1)? 
Happy to follow whatever the project's policy is.
Thanks for taking a look!

src/span_ext.rs
```diff
@@ -152,6 +157,63 @@ pub trait OpenTelemetrySpanExt {
...
+    /// ];
+    ///
+    /// // Add event using the extension method
+    /// app_root.add_otel_span_event("job_started".to_string(), dynamic_attrs);
+    ///
+    /// // ... perform work ...
+    ///
+    /// app_root.add_otel_span_event("job_completed", vec![KeyValue::new("status", "success")]);
+    /// ```
+    fn add_otel_span_event<T>(&self, name: T, attributes: Vec<KeyValue>)
```
djc, line 186: Are there other kinds of events in OTel? The name is a little verbose, so would be nice if we can get away with something a bit shorter like `add_otel_event()`.

djc, line 186: Nit: please use impl trait in arguments consistently where possible, so this should have `name: impl Into<Cow<'static, str>>` (everywhere).

examples/opentelemetry-span-event.rs
```diff
@@ -0,0 +1,65 @@
+use opentelemetry::trace::TracerProvider as _;
```
djc, line 1: This seems a little niche so I'd prefer to avoid maintaining an example for it.

src/span_ext.rs
```diff
@@ -237,4 +299,33 @@ impl OpenTelemetrySpanExt for tracing::Span {
...
+            let mut event = Some(opentelemetry::trace::Event::new(
+                name, timestamp, attributes, 0,
+            ));
+            if let Some(get_context) = subscriber.downcast_ref::<WithContext>() {
+                get_context.with_context(subscriber, id, move |data, _tracer| {
+                    if let Some(event) = event.take() {
+                        data.builder.events.get_or_insert_with(Vec::new).push(event);
+                    }
+                });
+            }
```
djc, line 328: Nit: I'm aware this is mostly copy/pasting, but please use let-else to reduce rightward drift where possible. (If you want to fix up the original function too, that would be nice, but please keep it in a separate commit.)

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Unreleased
 
+### Added
+
+- Add `OpenTelemetrySpanExt::add_otel_span_event` function to allow adding OpenTelemetry
+  events directly to a `tracing::Span`, enabling the use of dynamic attribute keys.
+- Add `OpenTelemetrySpanExt::add_otel_span_event_with_timestamp` function, similar to
+  `add_otel_span_event`, but allowing a specific `SystemTime` to be provided for the event.
```
djc, line 8: Nit: I think these changes belong in a single bullet point.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+- Add `examples/opentelemetry-span-event.rs` demonstrating the usage of the new span event methods.
```
djc, line 9: Don't forget to remove this when you remove the example.

djc (commented):
This is mostly looking good, thanks!

Please keep the clippy commits separate and squash any additional changes into your original commit (might be easier to reorder them so that the clippy commits come first).

src/span_ext.rs
alessandrobologna: Yes, there's a new [WIP spec for OTel Events](https://opentelemetry.io/docs/specs/semconv/general/events/), as a new signal type. More background about it in here: 

- https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0202-events-and-logs-api.md
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0265-event-vision.md
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md

I think though that it's still far away from being widely implemented and adopted, so I am happy with the span events API for now (and I don't necessarily agree that it was a good idea to introduce a new signal type, but that's digressing).

I agree that the name is verbose, but `add_otel_event` would then be a little ambiguous when those specs are more widely adopted and known. Alternatively, we could use `add_event` (as in the OTel Span trait, in the same way that you have `add_link`) and since this is a method of a trait that is meant to "wrap" an OTel span, maybe it would be clear that is not a Tracing event. Let me know your preference and I will update the PR.

examples/opentelemetry-span-event.rs
alessandrobologna: Ok I can remove it

CHANGELOG.md
alessandrobologna: will do

alessandrobologna:
@djc I think I have addressed all the changes requested. Let me know if you spot something else.

djc:
As mentioned:

> squash any additional changes into your original commit

Please squash the last two commits ("rename span event methods" and "use impl trait consistently"), which AFAICT only touch new code you have added, into your original commit ("add OpenTelemetrySpanExt methods").

alessandrobologna:
Hi @djc, thanks! I think I've addressed the feedback:

* Kept the clippy fixes separate and moved them first.
* Squashed the feature addition and subsequent refactorings into the final commit (78e8247).

The commit history should now match your suggestion. Let me know if there's something else!

djc:
You actually didn't address my feedback exactly, because you also squashed the refactoring that refactored other pre-existing code that you didn't add, so that's a little yucky from a reviewing perspective. However, I don't think it's worth doing another round trip on this, so I'll just merge this.

djc (approved):

alessandrobologna:
Apologies, I misunderstood, I just thought you wanted the clippy fixes separated. I see your point now.

Issue #199, opened 2025-04-22
OTLP Example logging with incorrect `service.name`


andreasWallnerIFX:
## Bug Report

Just a simple report on the OTLP example. I'm running the example as it is here: https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/opentelemetry-otlp.rs.

Logs to Jaeger will use the service name "unknown_service" instead of the crate name as intended.

### Version

<details>

<summary>version dumps</summary>

```
$ cargo tree | grep tracing
│   │   └── tracing v0.1.41
│   │       ├── tracing-attributes v0.1.28 (proc-macro)
│   │       └── tracing-core v0.1.33
│   │   │   └── tracing v0.1.41 (*)
│   │   └── tracing v0.1.41 (*)
│   │   └── tracing v0.1.41 (*)
│   └── tracing v0.1.41 (*)
│   │   ├── tracing-core v0.1.33 (*)
│   │   └── tracing-error v0.2.1
│   │       ├── tracing v0.1.41 (*)
│   │       └── tracing-subscriber v0.3.19
│   │           ├── tracing v0.1.41 (*)
│   │           ├── tracing-core v0.1.33 (*)
│   │           └── tracing-log v0.2.0
│   │               └── tracing-core v0.1.33 (*)
│   └── tracing-error v0.2.1 (*)
│   └── tracing v0.1.41 (*)
│   │   └── tracing v0.1.41 (*)
│   │   │   └── tracing v0.1.41 (*)
│   │       │   └── tracing v0.1.41 (*)
│   │       └── tracing v0.1.41 (*)
│   └── tracing v0.1.41 (*)
│   └── tracing v0.1.41 (*)
├── tracing v0.1.41 (*)
├── tracing-core v0.1.33 (*)
├── tracing-opentelemetry v0.30.0
│   ├── tracing v0.1.41 (*)
│   ├── tracing-core v0.1.33 (*)
│   ├── tracing-log v0.2.0 (*)
│   └── tracing-subscriber v0.3.19 (*)
├── tracing-subscriber v0.3.19 (*)
```

And the opentelemetry stuff:
```
$ cargo tree | grep opentelemetry
├── opentelemetry v0.29.1
├── opentelemetry-otlp v0.29.0
│   ├── opentelemetry v0.29.1 (*)
│   ├── opentelemetry-http v0.29.0
│   │   ├── opentelemetry v0.29.1 (*)
[...]

PR #203, opened 2025-05-06
OTLP Example logging with incorrect ‘service.name’

Dreistein:
## Motivation

Logs to Jaeger will use the service name "unknown_service" instead of the crate name as intended.

See #199 for details.

## Solution

Using `with_service_name` instead of declaring it in the Registry.


djc (approved):

andreasWallnerIFX:
@Dreistein  Just wanted to open this, thanks!

PR #210, opened 2025-06-05
chore(deps): upgrade pprof and remove obsolete forced minimum versions

mladedav:
## Motivation

Clean up our dependencies.

## Solution

`pprof` was the only thing that pulled `thiserror@1` so let's get rid of it. It's just a dev dependency.

The forced versions seem to have been fixed by upgrading `opentelemetry` and its transitive dependencies.


djc (approved):

mladedav:
I stand corrected, we still need the `lazy_static` one. Dev dependencies forced the correct version, not a transitive dependency.

PR #213, opened 2025-06-12
chore!: remove metrics feature gate

mladedav:
## Motivation

The upstream feature has been stabilized so there is no reason for use to hide the functionality behind an "unstable" feature.

## Solution

Remove the cargo feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,6 @@ rust-version = "1.75.0"
 default = ["tracing-log", "metrics"]
 # Enables support for exporting OpenTelemetry metrics
 metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
-# Enables experimental support for OpenTelemetry gauge metrics
-metrics_gauge_unstable = []
 
 [dependencies]
 opentelemetry = { version = "0.30.0", default-features = false, features = ["trace"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -19,8 +19,6 @@ rust-version = "1.75.0"
 default = ["tracing-log", "metrics"]
 # Enables support for exporting OpenTelemetry metrics
 metrics = ["opentelemetry/metrics","opentelemetry_sdk/metrics", "smallvec"]
-# Enables experimental support for OpenTelemetry gauge metrics
```
djc, line 22: Suggest adding a commit that makes a semver-incompatible bump to the version number in the same PR.

djc (approved):

Cargo.toml
mladedav: Personally, I'm not a fan of that because then I have to remember (or check) whether I have bumped the version between releases to not bump it again in another breaking change PR and to correctly bump or leave the patch version when a release does happen.

This way I have to just check what has happened since the last release once and if I correctly mark breaking changes with an `!`, then release-plz should catch it automatically.

Or do you have some other workflow that checks for you that you're not accidentally skipping versions?

djc: I'm not familiar with the `!` thing from release-plz -- if you have a good mechanism for tracking breaking changes, seems fine. (It was just a suggestion after all.)

PR #215, opened 2025-06-15
Fix: remove Lazy for thread_local static

Expyron:
As `thread_local` statics are already lazily initialized on first access, there is no need to use a Lazy initializer on top of that.

This PR removes that usage and the associated dependency on `once_cell`.


mladedav:
Thanks!

PR #218, opened 2025-07-02
Consolidates gauge metric filtering logic and add gauge example in docs

xinchengxx:
## Motivation
With gauge metrics now stable in upstream, we can simplify the filtering logic and improve documentation to better support gauge usage patterns.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -297,6 +297,9 @@ impl Visit for MetricVisitor<'_> {
 /// info!(histogram.qux = 1);
 /// info!(histogram.abc = -1);
 /// info!(histogram.def = 1.1);
+///
+/// info!(gauge.foo = 1);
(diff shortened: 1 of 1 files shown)


src/metrics.rs
```diff
@@ -297,6 +297,9 @@ impl Visit for MetricVisitor<'_> {
 /// info!(histogram.qux = 1);
 /// info!(histogram.abc = -1);
 /// info!(histogram.def = 1.1);
+///
+/// info!(gauge.foo = 1);
```
mladedav, line 302: Also add the prefix to the list just before this example please.

mladedav (changes requested):
Thanks!

src/metrics.rs
xinchengxx: Added~

mladedav (approved):

PR #224, opened 2025-08-06
fix(metrics)!: hold onto `MetricsProvider` in `MetricsLayer`

### LINKED ISSUE #178: MetricsLayer does not really pipe metrics to reader
## Bug Report
metrics reader never get anything from MetricsLayer

### Version

opentelemetry = { version = "0.26", features = ["trace"] }
opentelemetry-proto = { version = "0.26", features = ["trace"] }
opentelemetry_sdk = { version = "0.26", features = ["trace", "rt-tokio"] }
tracing-opentelemetry = "0.27"

### Description

image
image
image
image
image

mladedav:
## Motivation

Fixes #178 

Dropping `MeterProvider` early shuts down the `Meter` we carry aroud forcing users wither to keep a copy of the provider for the duration of their program or losing all their metrics.

## Solution

Keep the provider inside our layer. It adds a new generic argument to the type which is a breaking change.

The tests were written in a way that prevented this bug by having the provider cloned so I've got rid of that too.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -368,18 +368,19 @@ impl Visit for MetricVisitor<'_> {
 ///
 /// In the future, this can be improved by associating each `Metric` instance to
 /// its callsite, eliminating the need for any maps.
-///
 #[cfg_attr(docsrs, doc(cfg(feature = "metrics")))]
-pub struct MetricsLayer<S> {
+pub struct MetricsLayer<S, M> {
     inner: Filtered<InstrumentLayer, MetricsFilter, S>,
+    // We need to hold onto this so that the `InstrumentLayer` can use the created `Meter`.
+    _meter_provider: M,
 }
 
-impl<S> MetricsLayer<S>
+impl<S, M> MetricsLayer<S, M>
 where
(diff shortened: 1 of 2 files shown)


src/metrics.rs
```diff
@@ -467,7 +469,7 @@ where
     }
 }
 
-impl<S> Layer<S> for MetricsLayer<S>
+impl<S, M: 'static> Layer<S> for MetricsLayer<S, M>
```
djc, line 472: Why does this need the explicit lifetime?

djc (commented):
This looks like it'll be semver-breaking, maybe include a proactive semver-incompatible version bump?

src/metrics.rs
mladedav: Because `Layer<S>` has a `Self: 'static` clause.

mladedav:
As for the semver incompatible bump, I'm still planning to use `realease-plz` which bumps the version on release so that it's not accidentally bumped multiple times between releases. Since the commits contain conventional commits with an `!` it knows there have been breaking changes.

PR #225, opened 2025-09-05
fix(layer): use otel semantic conventions for code

PeterFaiman:
## Motivation

The otel semantic conventions specify the names `code.file.path` and `code.line.number` for these attributes. There is no equivalent for module, but I extrapolated from `code.function.name`.

https://opentelemetry.io/docs/specs/semconv/code/

## Solution

Rename the current attributes.

Happy to revert the change to `code.namespace`, since it's not actually in the spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+### Breaking Changes
+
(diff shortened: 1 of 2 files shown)


mladedav:
Thanks!

PR #229, opened 2025-09-25
chore: remove use of state field in SpanBuilder

bantonsson:
## Motivation

The state field will be removed from the public API of the OpenTelemetry `SpanBuilder` along with a number of other fields https://github.com/open-telemetry/opentelemetry-rust/issues/2886

This is the last of the removed fields being used.

## Solution

Keep the state in the data until the OpenTelemetry `Span` is created.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -163,7 +163,7 @@ struct SpanBuilderUpdates {
 }
 
 impl SpanBuilderUpdates {
-    fn update(self, span_builder: &mut SpanBuilder) {
+    fn update(self, span_builder: &mut SpanBuilder, s: &mut Status) {
         let Self {
             name,
             span_kind,
@@ -178,7 +178,7 @@ impl SpanBuilderUpdates {
             span_builder.span_kind = Some(span_kind);
         }
         if let Some(status) = status {
-            span_builder.status = status;
+            *s = status;
         }
         if let Some(attributes) = attributes {
             if let Some(builder_attributes) = &mut span_builder.attributes {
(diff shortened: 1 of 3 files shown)


tests/span_ext.rs
```diff
@@ -171,15 +171,15 @@ fn test_add_link_variants() {
     let (_tracer, provider, exporter, subscriber) = test_tracer();
 
     let link_builder_cx = opentelemetry::trace::SpanContext::new(
-        opentelemetry::trace::TraceId::from_u128(0x1234567890abcdef1234567890abcdef),
-        opentelemetry::trace::SpanId::from_u64(0x1234567890abcdef),
```
bantonsson: These will be removed as well in upcoming releases.

mladedav:
Ah, sorry, I forgot about this during the day but we can still do this in a point release.

That said, I'd also like to untangle the dependencies between opentelemtry and us because now we're in a weird state where presumably their appender depends on an older version without the context propagation rework and will continue to until the next breaking release.

bantonsson:
No worries. Doing this in a point release is fine.

Yes, moving the appenders to https://github.com/open-telemetry/opentelemetry-rust-contrib would be a good thing to break up the circularity.

src/layer.rs
```diff
@@ -1358,12 +1375,19 @@ where
...
+                    status,
+                } => {
                     // Don't create the context here just to get a SpanRef since it's costly
                     let mut span = builder.start_with_context(&self.tracer, &parent_cx);
                     if let Some(timings) = timings {
                         span.set_attributes(timings)
                     };
+                    if status != Status::Unset {
+                        span.set_status(status);
+                    }
```
mladedav: Since the user (or another library) can also set this manually through opentelemtry, I don't think we should  overwrite this. Especially if they've explicitly set ok and we would change it back to error.

On the other hand, the otel spec says that trying to set status to unset should be ignored and once the status is set to ok, other attempts to set the status should be ignored, so I think that if that is implemented correctly, we can just unconditionally call `set_status` here and simplify the code that way.

src/layer.rs
```diff
@@ -974,8 +974,16 @@ where
...
+        } else if let OtelDataState::Builder {
+            builder,
+            parent_cx,
+            status,
+        } = take(&mut otel_data.state)
+        {
+            let mut span = builder.start_with_context(&self.tracer, &parent_cx);
+            if status != Status::Unset {
+                span.set_status(status);
+            }
```
mladedav: Let's make this unconditional, if the condition doesn't hold, it's just a noop.

mladedav: You did not change this?

bantonsson: Hey, looks like I messed up. Things should be changed now.

mladedav (approved):
Thanks!

PR #230, opened 2025-09-26
Upgrade to otel 0.31

tisonkun:
Minor changes.

cc @djc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - The attributes `code.filepath`, `code.lineno`, and `code.namespace` have been renamed to `code.file.path`, and
   `code.line.number`, and `code.module.name`, to align with the opentelemetry semantic conventions for code.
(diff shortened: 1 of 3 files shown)


emilk:
I would love to have a quick release of this, as it is blocking out update of `opentelemetry` and `tonic` 🙏

djc:
> cc @djc

There is no need to ping me personally.

> I would love to have a quick release of this, as it is blocking out update of `opentelemetry` and `tonic` 🙏

And this is the case every time -- please don't pile on within a few hours.

tisonkun:
> There is no need to ping me personally.

Got it.

mladedav (approved):
Next time, please don't do unnecessary formatting changes in the whole file when you're just bumping a few versions.

But thanks for the fast PR.

tisonkun:
> Next time, please don't do unnecessary formatting changes in the whole file when you're just bumping a few versions.

Got it.

> But thanks for the fast PR.

🎉 Then downstream projects are waiting for a new release :D

Issue #227, opened 2025-09-20
Change visibility of `OtelData` to `pub`


ollyswanson:
### Motivation

A recently merged PR https://github.com/tokio-rs/tracing-opentelemetry/pull/202 has changed the visibility of `OtelData`. It was previously `pub` and it was possible to access it from other layers by poking around in the span extensions - I've been using this to inject trace_id and span_id into my logs.

This is also being used by the [opentelemetry-appender-tracing](https://github.com/open-telemetry/opentelemetry-rust/blob/5250df2f419b43c4eef0fca4ffb04edd943c61b6/opentelemetry-appender-tracing/src/layer.rs#L249-L266) crate for, I think, the same purpose.

### Proposal

Make [`OtelData`](https://github.com/tokio-rs/tracing-opentelemetry/blob/4fdf56048d1e710a3fd2164a57ba8ce3b7bd10d4/src/lib.rs#L133-L138) pub again as it was [previously](https://github.com/tokio-rs/tracing-opentelemetry/blob/e43bb49706f60b2ecde38d7645be3142bdd5a2f2/src/lib.rs#L134-L140).

I think it might be simpler to just make it `pub` again as I can't discern any motivation for changing the visibility. I'm happy to raise a PR for it but I want to understand if there was a motivation for the change before doing that.

### Alternatives

If there's a better way of doing this I'd be happy to adopt that pattern instead, but up until now it's been a very convenient way to transparently include the trace_id and span_id in my own logging layer.

mladedav:
Thanks for catching this.

I'm pretty sure we wouldn't want to make it structurally public (it was bad enough before, now unexpected changes to the state would probably be even worse), but we should definitely expose the IDs through a method.

I don't think there really is another way because for example `tracing` events can be emitted even when their parent span is not entered so this should not rely on the active `opentelemetry` context.

ollyswanson:
Exposing them through a method would be lovely - I'm up to my eyeballs with work at the moment but I could try to get a PR in in the next couple of weeks.

Don't consider that to be me asking to squat on it though if someone else gets there first it's one less thing to have to worry about 😄

PR #232, opened 2025-09-29
chore: prepare release of 0.32.0

mladedav:
I had the changelog autogenerated from commits and then cleaned it up. It also added a header to the changelog and changed the format (e.g. date for the release) but that should be fine.

After merging, I'll release to crates.io and add git tags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,41 @@
-# Unreleased
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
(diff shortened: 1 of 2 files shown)


djc (approved):

cschramm:
The removal of `OtelData` from the API is probably worth mentioning. #202 removed it and use cases as the one described in #227 will not be able to just update to this release.

mladedav:
You're right that this is something we should have in the release. If we merge it, I'll edit the changelog before merging.

bantonsson:
Can we pretty please have #229 merged before the release? It is necessary for cleanup work in OTel.

mladedav:
I thought it's in opentelemtry itself? As long as we merge it before the next release of opentelemtry it doesn't matter.

As opentelemtry doesn't depend on this repo, I don't see how having that in this or the next release changes anything but maybe I'm missing some dependency?

bantonsson:
The OpenTelemetry repo unfortunately has a tracing logging appender that depends on this repo. It’s not one of the core packages, but it makes releasing (and merging PRs) complicated because of the circular nature.

PR #233, opened 2025-09-29
feat: make trace ID and span ID public on `OtelData`

### LINKED ISSUE #227: Change visibility of `OtelData` to `pub`

mladedav:
## Motivation

Get the IDs from the `OtelData` struct.

Closes #227

## Solution

When the context is started, we can get the IDs.

There is however nothing to report when the context is not yet fully built. But if it is not, there is not span ID to report and reporting the parent span's trace ID from the builder may not be correct as that may still change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,15 +126,50 @@ pub use layer::{layer, OpenTelemetryLayer};
 
 #[cfg(feature = "metrics")]
 pub use metrics::MetricsLayer;
+use opentelemetry::trace::TraceContextExt as _;
 pub use span_ext::OpenTelemetrySpanExt;
 
 /// Per-span OpenTelemetry data tracked by this crate.
 #[derive(Debug)]
-pub(crate) struct OtelData {
+pub struct OtelData {
     /// The state of the OtelData, which can either be a builder or a context.
     state: OtelDataState,
     /// The end time of the span if it has been exited.
     end_time: Option<SystemTime>,
+}
+
+impl OtelData {
+    /// Gets the trace ID of the span.
+    ///
+    /// Returns `None` if the context has not been built yet. This can be forced e.g. by calling
+    /// [`context`] on the span (not on `OtelData`) or if [context activation] was not explicitly
+    /// opted-out of, simply entering the span for the first time.
+    ///
+    /// [`context`]: OpenTelemetrySpanExt::context
+    /// [context activation]: OpenTelemetryLayer::with_context_activation
+    pub fn trace_id(&self) -> Option<opentelemetry::TraceId> {
+        if let OtelDataState::Context { current_cx } = &self.state {
+            Some(current_cx.span().span_context().trace_id())
+        } else {
+            None
+        }
+    }
+
+    /// Gets the span ID of the span.
+    ///
+    /// Returns `None` if the context has not been built yet. This can be forced e.g. by calling
+    /// [`context`] on the span (not on `OtelData`) or if [context activation] was not explicitly
+    /// opted-out of, simply entering the span for the first time.
+    ///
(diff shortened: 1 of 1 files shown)


mladedav:
cc @cschramm and @ollyswanson if you think this is good enough.

What wouldn't specifically work is something like
```rust
let span = tracing::debug_span!("root");
tracing::error!(parent: &span, " :( ");
```
while starting the span would work
```rust
let span = tracing::debug_span!("root");
let _guard = span.entered();
tracing::error!(" :) ");
```

cschramm:
Works perfectly fine for my use case, where I use the dispatcher's current span (or rather its root span, by traversing the parents). If I'm not mistaken, that should always be entered.

mladedav:
> If I'm not mistaken, that should always be entered.

If you're talking about `on_enter` then yes, if you're talking about `on_event`, then no, see the first example of what would be broken if I understand what you mean.

But generally, people tend to use enter spans rather than explicitly setting the parent so you should be mostly fine.

bantonsson:
Hey @mladedav I'm working on making it possible to get the `context` from the `SpanRef` instead of making the `OtelData` public again. I ran into this bug https://github.com/tokio-rs/tracing/pull/3379, but I have a workaround for that. It would ensure that the `context` gets built when it is requested. I think that would be useful for other things like `baggage`.

src/lib.rs
```diff
@@ -126,17 +126,48 @@ pub use layer::{layer, OpenTelemetryLayer};
...
     state: OtelDataState,
     /// The end time of the span if it has been exited.
     end_time: Option<SystemTime>,
 }
 
+impl OtelData {
+    /// Gets the trace ID of the span.
+    ///
+    /// Returns `None` if the context has not been built yet. This can be forced e.g. by calling
+    /// [`context`] on the span (not on `OtelData`) or simply entering the span for the first time.
```
bantonsson: If you disable the `context_activation` the context won't be created when you enter the span.

bantonsson:
I opened #234 as another solution

mladedav:
For simplicity, I'd go with making it public again. We can deprecate this in the next version (and add the context activation-based approach in a point release whenever we want).

src/lib.rs
mladedav: Thanks, fixed.

bantonsson (approved):

Issue #121, opened 2024-04-04
WithContext not populated when a reload layer is used


arlyon:
## Bug Report

Hi there. I am building a cli which may-or-may-not have opentelemetry enabled. We initialize tracing early on and conditionally enable opentelemetry based on some flags. We use reload layers starting with `None` to allow enabling layers later on. Everything works, except context propagation. If the layer is not present when the registry is init'ed, `WithContext` is not set up correctly meaning `set_parent` and friends fails when attempting to downcast the subscriber.

https://github.com/tokio-rs/tracing-opentelemetry/blob/5e3354ec24debcfbf856bfd1eb7022459dca1e6a/src/span_ext.rs#L139-L150

I can confirm this _does_ work as expected even with a reload layer as long as the layer starts as `Some(T)`. When the layer is None at `init` time it does not. I would expect when reloading the layer that the subscriber is configured appropriately such that span propagation is possible.

### Version

```
│   │   │   │   └── tracing v0.1.37
│   │   │   │       ├── tracing-attributes v0.1.26 (proc-macro)
│   │   │   │       └── tracing-core v0.1.31
│   │   │   └── tracing v0.1.37 (*)
│   │   ├── tracing v0.1.37 (*)
│   │   └── tracing v0.1.37 (*)
│   │   │   └── tracing v0.1.37 (*)
│   │   ├── tracing v0.1.37 (*)
│   └── tracing v0.1.37 (*)
│   ├── tracing v0.1.37 (*)
│   ├── tracing v0.1.37 (*)
│   │       └── tracing v0.1.37 (*)
├── tracing v0.1.37 (*)
├── tracing-appender v0.2.2
│   └── tracing-subscriber v0.3.18
│       ├── tracing v0.1.37 (*)
│       ├── tracing-core v0.1.31 (*)
│       └── tracing-log v0.2.0
│           └── tracing-core v0.1.31 (*)
├── tracing-chrome v0.7.1
│   ├── tracing-core v0.1.31 (*)
│   └── tracing-subscriber v0.3.18 (*)
├── tracing-opentelemetry v0.23.0
│   ├── tracing v0.1.37 (*)
│   ├── tracing-core v0.1.31 (*)
│   ├── tracing-log v0.2.0 (*)
│   └── tracing-subscriber v0.3.18 (*)
├── tracing-subscriber v0.3.18 (*)
│   ├── tracing v0.1.37 (*)
│   │   ├── tracing v0.1.37 (*)
[...]

PR #234, opened 2025-09-30
feat: allow OpenTelemetry context access from SpanRef

bantonsson:
## Motivation

Since `OtelData` is no longer public, there needs to be a way to access the OpenTelemetry context from other Layers and ensure that the context is created.

## Solution

Introduce `OpenTelemetryContext` that given a `SpanRef` and the `Dispatch` for the layers will get or create the OpenTelemetry context.

## Note

Both the test and the example has a workaround for https://github.com/tokio-rs/tracing/pull/3379 which makes the code unnecessarily complicated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ opentelemetry = { version = "0.31.0", default-features = false, features = [
 
 tracing = { version = "0.1.35", default-features = false, features = ["std"] }
 tracing-core = "0.1.28"
-tracing-subscriber = { version = "0.3.0", default-features = false, features = [
+tracing-subscriber = { version = "0.3.22", default-features = false, features = [
     "registry",
     "std",
 ] }
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -19,7 +19,7 @@ use tracing_core::{field, Event, Subscriber};
 use tracing_log::NormalizeEvent;
 use tracing_subscriber::layer::Context;
 use tracing_subscriber::layer::Filter;
-use tracing_subscriber::registry::LookupSpan;
+use tracing_subscriber::registry::{ExtensionsMut, LookupSpan};
 use tracing_subscriber::Layer;
 #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
 use web_time::Instant;
@@ -118,6 +118,14 @@ pub(crate) struct WithContext {
     #[allow(clippy::type_complexity)]
     pub(crate) with_activated_context:
         fn(&tracing::Dispatch, &span::Id, f: &mut dyn FnMut(&mut OtelData)),
+
+    ///
+    /// Ensures the given SpanId has been activated - that is, created in the OTel side of things,
+    /// and had its SpanBuilder consumed - and then provides access to the OtelData associated with it.
+    ///
+    #[allow(clippy::type_complexity)]
+    pub(crate) with_activated_context_extensions:
+        fn(&tracing::Dispatch, &mut ExtensionsMut<'_>, f: &mut dyn FnMut(&mut OtelData)),
 }
 
 impl WithContext {
@@ -146,6 +154,20 @@ impl WithContext {
         mut f: impl FnMut(&mut OtelData),
     ) {
         (self.with_activated_context)(dispatch, id, &mut f)
+    }
+
+    ///
+    /// Ensures the given SpanId has been activated - that is, created in the OTel side of things,
+    /// and had its SpanBuilder consumed - and then provides access to the OtelData associated with it.
+    ///
+    #[allow(clippy::type_complexity)]
+    pub(crate) fn with_activated_context_extensions(
+        &self,
+        dispatch: &tracing::Dispatch,
+        extensions: &mut ExtensionsMut<'_>,
+        mut f: impl FnMut(&mut OtelData),
+    ) {
+        (self.with_activated_context_extensions)(dispatch, extensions, &mut f)
     }
 }
 
@@ -639,6 +661,7 @@ where
             with_context: WithContext {
                 with_context: Self::get_context,
                 with_activated_context: Self::get_activated_context,
+                with_activated_context_extensions: Self::get_activated_context_extensions,
             },
             _registry: marker::PhantomData,
         }
@@ -694,6 +717,8 @@ where
             with_context: WithContext {
                 with_context: OpenTelemetryLayer::<S, Tracer>::get_context,
                 with_activated_context: OpenTelemetryLayer::<S, Tracer>::get_activated_context,
+                with_activated_context_extensions:
+                    OpenTelemetryLayer::<S, Tracer>::get_activated_context_extensions,
             },
             _registry: self._registry,
             // cannot use ``..self` here due to different generics
@@ -963,11 +988,20 @@ where
             .span(id)
             .expect("registry should have a span for the current ID");
 
+        let mut extensions = span.extensions_mut();
+
+        Self::get_activated_context_extensions(dispatch, &mut extensions, f)
+    }
+
+    fn get_activated_context_extensions(
+        dispatch: &tracing::Dispatch,
+        extensions: &mut ExtensionsMut<'_>,
+        f: &mut dyn FnMut(&mut OtelData),
+    ) {
         let layer = dispatch
             .downcast_ref::<OpenTelemetryLayer<S, T>>()
             .expect("layer should downcast to expected type; this is a bug!");
 
-        let mut extensions = span.extensions_mut();
         if let Some(otel_data) = extensions.get_mut::<OtelData>() {
             // Activate the context
             layer.start_cx(otel_data);
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -121,11 +121,11 @@ pub(crate) struct WithContext {
 
     ///
     /// Ensures the given SpanId has been activated - that is, created in the OTel side of things,
-    /// and had its SpanBuilder consumed - and then provides access to the OtelData associated with it.
+    /// and had its SpanBuilder consumed - and then provides access to the OTel Context associated with it.
     ///
     #[allow(clippy::type_complexity)]
-    pub(crate) with_activated_context_extensions:
-        fn(&tracing::Dispatch, &mut ExtensionsMut<'_>, f: &mut dyn FnMut(&mut OtelData)),
+    pub(crate) with_activated_otel_context:
+        fn(&tracing::Dispatch, &mut ExtensionsMut<'_>, f: &mut dyn FnMut(&OtelContext)),
 }
 
 impl WithContext {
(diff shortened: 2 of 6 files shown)


mladedav:
I'll have a closer look later, but on first look, I have some reservations. But maybe they'll be solvable in this design.

The first thing I noticed when skimming this is that `SpanRefExt::context` should definitely take `&mut self` because it internally takes an exclusive lock on its extensions which would definitely lead to people deadlocking themselves through that.

I don't think this makes sense to rush this now so I'll leave it out of today's release and go with the simpler version that does not start the context automatically.

bantonsson:
Thanks for noticing @mladedav. That is definitely a problem. No need to rush it.

I don't think that a `&mut self` will protect against the locking issue. I think that the `ExtensionMut` needs to be passed into the function to make it explicit.

mladedav:
You're right, for some reason I though when you have `ExtensionMut` it mutably borrows the `SpanRef` but it's just normal borrow (and on top of that you can just get another `SpanRef` by callign `ctx.span(&span.id)`).

Passing the extensions explicitly is also probably better since the user will most likely already have the lock or will want to get it later.

bantonsson:
@mladedav I've reworked this a bit so the `ExtensionsMut` is now passed in explicitly.

bantonsson:
@mladedav I think this is ready for review whenever you have the time.

src/otel_context.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// }
+    /// ```
+    ///
+    /// # Use Cases
+    ///
+    /// - When working with multiple subscriber configurations
+    /// - When implementing advanced tracing middleware that manages multiple dispatches
+    pub fn context(
+        extensions: &mut ExtensionsMut<'_>,
+        dispatch: &Option<Dispatch>,
```
mladedav: Can we just require `&Dispatch`? It seems that if it's `None`, we return `None` so I don't think there is a reason to take an `Option`.

bantonsson:
There seems to be a release coming soon for `tracing-subscriber` with the fix for [`on_register_dispatch`](https://github.com/tokio-rs/tracing/pull/3379) that would clean up the tests and examples.

src/layer.rs
```diff
@@ -954,11 +979,20 @@ where
...
+    }
+
+    fn get_activated_context_extensions(
+        dispatch: &tracing::Dispatch,
+        extensions: &mut ExtensionsMut<'_>,
+        f: &mut dyn FnMut(&mut OtelData),
+    ) {
         let layer = dispatch
             .downcast_ref::<OpenTelemetryLayer<S, T>>()
             .expect("layer should downcast to expected type; this is a bug!");
```
mladedav, line 1043: This comment is more for my future reference. The behavior was here before this PR, so you don't have to do anything about it now.

I think this can panic if we a layer with bad downcasting implementation such as `reload::Layer`.

src/layer.rs
```diff
@@ -954,11 +979,20 @@ where
...
+        let mut extensions = span.extensions_mut();
+
+        Self::get_activated_context_extensions(dispatch, &mut extensions, f)
+    }
+
+    fn get_activated_context_extensions(
+        dispatch: &tracing::Dispatch,
+        extensions: &mut ExtensionsMut<'_>,
+        f: &mut dyn FnMut(&mut OtelData),
+    ) {
```
mladedav, line 1040: Can `f` directly take `opentelemetry::Context`?

Right now, we give a wrapper, that contains an end time (if someone wants to change that, they can directly get the otel data, no need to do anything around context activation) and a state which we pretty much guarantee to contain a case with `opentelemetry::Context` inside.

This then leads to stuff like
```
                    |data: &mut OtelData| {
                        if let OtelDataState::Context { current_cx } = &data.state {
                            cx = Some(current_cx.clone());
                        }
                    },
```
which is the `OpenTelemetryContext::context` implementation and is a bit weird considering we just made sure the context was started.

src/otel_context.rs
```diff
@@ -0,0 +1,72 @@
...
+use tracing::Dispatch;
+use tracing_subscriber::registry::ExtensionsMut;
+
+/// Utility functions to allow tracing [`ExtensionsMut`]s to return
+/// [OpenTelemetry] [`Context`]s.
+///
+/// [`ExtensionsMut`]: tracing_subscriber::registry::ExtensionsMut
+/// [OpenTelemetry]: https://opentelemetry.io
+/// [`Context`]: opentelemetry::Context
+pub struct OpenTelemetryContext {}
```
mladedav: We don't need this struct, we can just provide a free-standing function (`get_otel_context` or whatever).

---

If we really wanted a struct for some reason, we'd probably want to make it non-constructible like `enum OpenTelemetryContext {}`.

src/layer.rs
mladedav: We need to do that since we'll probably want to hide `OtelData` again in the next breaking release. We made it public again just so people got access to trace/span IDs inside tracing layers and with this PR getting just the `Context` is sufficient.

src/layer.rs
mladedav: This is already tracked in #121 

This is probably fine because to get here, we already had to downcast the subscriber down to `WithContext` so this is going to work.

src/otel_context.rs
bantonsson: This is just convenience. The `WeakDispatch::upgrade()` method returns an `&Option<Dispatch>` since the underlying `Dispatch` could have been dropped. That shouldn't happen if this is used from layers that are stacked together, so requiring the end user to check it seems unnecessary.

src/layer.rs
bantonsson: All these things are `pub(crate)` so `OtelData` doesn't leak, but I can change the signature.

src/otel_context.rs
bantonsson: I'll remove the struct.

Cargo.toml
```diff
@@ -29,7 +29,8 @@ opentelemetry_sdk = { version = "0.31.0", default-features = false, features = [
 ] }
 tracing = { version = "0.1.35", default-features = false, features = ["std"] }
 tracing-core = "0.1.28"
-tracing-subscriber = { version = "0.3.0", default-features = false, features = [
+# The 0.3.22 version contains the fix for on_register_dispatch in Layer
```
mladedav: Don't include the comment, it may be useful for reviewing, but I don't see a reason to keep this kind of information in `Cargo.toml`.

src/layer.rs
mladedav: Please do, it still doesn't make sense that the provided callback needs to match over `data.state` when the whole point is guaranteeing that there is a `context`.

src/otel_context.rs
mladedav: It returns an owned `Option<Dispatch>`. Taking at least `Option<&Dispatch>` with `dispatch.upgarde().as_ref()` instead of `&dispatch.upgarde()` would be slightly better ([clippy lint with rationale](https://rust-lang.github.io/rust-clippy/master/index.html?search=ref_optio#ref_option)), but I still think that this should not take an `Option` if we don't intend to do anything with that.

I don't think it's unreasonable to require a value that we really need.

src/layer.rs
bantonsson: It’s already changed. That method is now only used internally.

src/otel_context.rs
bantonsson: Ok. I'll change the API to not take an `Option`.

Cargo.toml
bantonsson: Ok, I'll remove the comment.

src/layer.rs
mladedav: I don't know if we're talking about different things but I still see `f: &mut dyn FnMut(&mut OtelData),` here.

bantonsson: Sorry for not being more clear. This method is internal to the `layer`. It's called from both `get_activated_context` and `get_activated_otel_context` that are the _exposed_ methods that are used from outside the `layer`.

This is the signature that is used for the otel context:

```rust
    fn get_activated_otel_context(
        dispatch: &tracing::Dispatch,
        extensions: &mut ExtensionsMut<'_>,
        f: &mut dyn FnMut(&OtelContext),
    ) {
...
    }
```

mladedav:
Thanks for pushing this through and sorry for the sporadic reviews.

pronebird:
This PR is too hard to understand. I used to grab extensions directly and then extract trace_id/span_id:

```rs
let Some((trace_id, span_id)) = ctx.event_scope().and_then(|mut scope| {
  scope.find_map(|span_ref| {
    let exts = span_ref.extensions();
    let otel = exts.get::<OtelData>()?;
    Some((otel.trace_id()?.to_string(), otel.span_id()?.to_string()))
  })
})
```

But this approach no longer works. What's the migration path?

Does this look correct?

```rs
let Some((trace_id, span_id)) = ctx.event_scope().and_then(|mut scope| {
    tracing::dispatcher::get_default(|dispatch| {
        scope.find_map(|span_ref| {
            let otel = get_otel_context(&span_ref.id(), dispatch)?;
            let span = otel.span();
            let span_ctx = span.span_context();

            let trace_id = span_ctx.trace_id();
            let span_id = span_ctx.span_id();

            Some((trace_id.to_string(), span_id.to_string()))
        })
    })
})
```

mladedav:
See [this example](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/otel_context.rs).

The most relevant parts are [this](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/otel_context.rs#L86-L95) and [that](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/otel_context.rs#L71).

What you're doing seems correct to me on the first look but I didn't check closer.

pronebird:
Thanks for pointers. I am implementing the following trait:

```rs
impl<S, N> FormatEvent<S, N> for JsonLogFormatter
where
    S: Subscriber + for<'a> LookupSpan<'a>,
    N: for<'a> FormatFields<'a> + 'static,
```

There is no dispatch and `Subscriber` is buried inside of `Context` passed to:

```rs
fn format_event(
        &self,
        ctx: &FmtContext<'_, S, N>,
        mut writer: tracing_subscriber::fmt::format::Writer<'_>,
        event: &Event<'_>,
    ) -> std::fmt::Result
```

Hence I am at loss. I don't understand how to use `get_otel_context`. The fact that the code compiles doesn't give me assurance that it's correct as I think `tracing::dispatcher::get_default` is probably the wrong call to obtain `Dispatch`.

More importantly, why can't there be a simpler call that doesn't require obtaining `Dispatch`. Everything used to work perfectly fine by working directly with `SpanRef` and extensions until `OtelData` was made private and replaced with a method that lacks ergonomics and demands types that aren't even available in my context:

```rs
let exts = span_ref.extensions();
let otel = exts.get::<OtelData>()?;
```

mladedav:
So you're doing a custom format for the `tracing-subscriber`'s `fmt` layer, right?

I don't think that's easily done right now. I think that getting the `Dispatch` the way you're hinting would panic because it's in a `RefCell` and you're most likely already inside a scope where you had to dispatch the event you're handling. At least that's the case for the global dispatch I think.

Your options as I see them right now are:
- use something else for the JSON output like `json-subscriber`
- use an older OpenTelemetry version that works with this
- smuggle the dispatch into the format struct. You could use `Arc<OnceLock<WeakDispatch>>` and then seed that from a custom layer that only handles `on_register_dispatch`.

But I can appreciate that this is hard to use as is. I might look into options to expose the IDs better, e.g. by reintroducing extensions to expose the information when it's available and when it wouldn't be.

I'd just like to note that the point of this change wasn't to make the extensions private but to make automatic activation of OpenTelemetry contexts work better so that there is interoperability with upstream.

pronebird:
Thanks for taking time to write that down. That helped a lot! 🎉 

I went ahead and implemented a generic `Layer` capturing and sharing `Arc<OnceLock<WeakDispatch>>` among two of my layers. One of findings was that registration call wouldn't happen for boxed `Layer` 🤦

mladedav:
I think boxed layer should work on latest `tracing-subscriber` but e.g. `Vec<Box<syn Subscriber>>` won't because of the vector (but it's being fixed).

PR #237, opened 2025-10-13
Export `SetParentError`

### LINKED ISSUE #236: `SetParentError` not exported
### Version

0.32.0

### Description

I'm calling `set_parent()` on my span. I have a `From` implementation which maps the various errors I could encounter to `Response`s which have the appropriate HTTP status codes. It's not possible (as far as I can tell) to import `SetParentError` to set up a mapping for this error.

Could it be exported please?

mladedav:
## Motivation

Closes #236

The error was accidentally not publicly exported.

## Solution

Export the error.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,9 @@ web-time = "1.0.0"
 [lib]
 bench = false
 
+[lints.rust]
+unnameable_types = "warn"
+
 [[bench]]
 name = "trace"
 harness = false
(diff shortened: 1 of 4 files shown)


djc:
Consider setting `unnameable_types` to `warn` (I forget if this is clippy or the compiler).

mladedav:
Thanks, I've added it. I didn't know this one even existed.

I'll add some more lints in a separate PR.

djc (approved):

PR #238, opened 2025-10-19
chore: remove thiserror and unused dependencies

CathalMullan:
## Motivation

Was aiming to reduce `syn` usage in my dependency tree, and noticed a fairly unnecessary use of `thiserror` here. Also noticed a handful of unused deps, thanks to `cargo-shear`.

## Solution

Removal of the unused deps, and replacing `thiserror` with a manual implementation.


djc (approved):

mladedav (approved):
`opentelemetry` (which is a dependency) uses `thiserror` so I don't think there is any real benefit in doing this manually.

But it's true it's not much code so I'm fine with this.

Thanks for the pruning.

PR #243, opened 2025-12-05
Add exception.type to error events

### LINKED ISSUE #242: #[instrument(err)] causes broken traces in Datadog
I've been sent here after having a conversation in CNCF Slack in OTel Rust group. The issue tokio-rs/tracing#2648 was reported before, but it doesn't seem like that's a correct place for it.

## Bug Report
We are using the `tracing` framework for our Rust applications for distributed tracing and sending traces and spans to Datadog using OTLP+gRPC.

When using `#[tracing::instrument(err)]` macro to capture and record `Result::Err()` returned by the annotated function, traces end up being broken in Datadog, reporting: `Missing error message and stack trace`.

I am willing to work through this issue and contribute a fix, provided some guidance as to where do I start from.

Please let me know if you need any additional information here.

### Version
```
tracing v0.1.43
tracing-attributes v0.1.31 (proc-macro)
tracing-core v0.1.35
tracing-opentelemetry v0.32.0
tracing-subscriber v0.3.22
```

### Platform
Google's Distroless container images based on Debian 12, x86_64.

### Description
Here's the minimal reproducible example:
```rust
use opentelemetry::trace::TracerProvider as _;
use opentelemetry_otlp::WithExportConfig as _;
use tracing_subscriber::{layer::SubscriberExt as _, util::SubscriberInitExt as _};

#[tracing::instrument("Failing span", err)]
fn failing_function() -> Result<(), std::io::Error> {
    Err(std::io::Error::other("Big error, much problem"))
}

#[tokio::main]
async fn main() {
    let span_exporter = opentelemetry_otlp::SpanExporter::builder()
        .with_tonic()
        .with_protocol(opentelemetry_otlp::Protocol::Grpc)
        // Update to point to your collector
        .with_endpoint("grpc://127.0.0.1:4317")
        .with_timeout(std::time::Duration::from_secs(5))
        .build()
        .expect("Failed to build a span exporter");

    let tracer_provider = opentelemetry_sdk::trace::SdkTracerProvider::builder()
        .with_batch_exporter(span_exporter)
        .build();

    let traces_layer =
        tracing_opentelemetry::OpenTelemetryLayer::new(tracer_provider.tracer("tracing-repro"));

    tracing_subscriber::registry()
        .with(traces_layer)
        .try_init()
        .expect("Failed to initialize tracing");

    failing_function().expect_err("Expected failing function to fail");

    tracer_provider
        .shutdown_with_timeout(std::time::Duration::from_secs(5))
        .expect("Failed to shutdown the tracer provider");
}
```
and the `Cargo.toml` to go with it:
```toml
[package]
name = "tracing-instrument-otel-err"
version = "0.1.0"
edition = "2024"

[dependencies]
opentelemetry = { version = "0.31.0", default-features = false, features = ["trace"] }
opentelemetry-otlp = { version = "0.31.0", default-features = false, features = ["grpc-tonic", "trace"] }
opentelemetry_sdk = { version = "0.31.0", default-features = false, features = ["trace"] }
tokio = { version = "1.48.0", default-features = false, features = [
	"macros",
	"rt-multi-thread",
] }
tracing = { version = "0.1.43", default-features = false, features = ["attributes", "std"] }
tracing-opentelemetry = { version = "0.32.0", default-features = false }
tracing-subscriber = { version = "0.3.22", default-features = false, features = ["std", "registry"] }
```
When using this code and sending traces to Datadog, the following problem can be observed:

Image

and the error tab, respectively:

Image

### LINKED ISSUE #2648: `#![tracing::instrument(err)]` doesn't work with opentelemetry / datadog
### Version

Applies to the latest released versions as of this writing:
* `tracing = "0.1.37"`
* `tracing-attributes = "0.1.26"`
* `tracing-opentelemetry = "0.19.0"`
* `opentelemetry = "0.19.0"`
* `opentelemetry-datadog = "0.7.0"`
* `opentelemetry-otlp = "0.12.0"`
* ...

### Platform

any

### Crates

Interaction between `tracing-attributes`, `tracing-opentelemetry`, `opentelemetry`, and `opentelemetry-datadog`.

### Description

I'd like the full integration from `#[tracing(err)]` to just work, resulting in following the expected protocols:

* OpenTelemetry [says](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md) that spans should report an error by having an event named `exception` with attributes `exception.escaped` (which should always be `true` the case of `#[tracing::instrument(err)]`), `exception.message`, `exception.type`, `exception.stack`.
* Datadog [says](https://docs.datadoghq.com/tracing/trace_explorer/trace_view/?tab=spantags#more-information) that there should be attributes on the span: `error.msg`, `error.type`, `error.stack`.

From observation: it doesn't work with `opentelemetry_datadog`. E.g., if I use essentially `tracing_subscriber::registry().with(tracing_opentelemetry::layer().with_tracer(opentelemetry_datadog::new_pipeline().install_batch(...))`, Datadog will correctly identify that the span has an error, but I can't see any details anywhere. I just see `Missing error message and stack trace` from the Datadog UI.

From code inspection: I don't think it'd work well with `opentelemetry_otlp` either. In fact, I think `tracing-attributes` and `tracing-opentelemetry` aren't following the same protocol:

* `#[tracing(err)]` code [here](https://github.com/tokio-rs/tracing/blob/27f688efb72316a26f3ec1f952c82626692c08ff/tracing-attributes/src/expand.rs#L243) will create an event (with no message, I think?) with attribute `error` that has the `Display` or `Debug` output
* `tracing_opentelemetry` [has](https://github.com/tokio-rs/tracing-opentelemetry/blob/42e986929080a5f1f245876c1aaa82ce9a7d85de/src/layer.rs#L213) this promising-looking logic that looks based on the OpenTelemetry semantic conventions. It will fill in `exception.message` and `exception.stack`...but only on `record_error`. So the conversion to `Display` or `Debug` (`error = %e` or `error = ?e`) defeats that by calling `record_debug` instead in both cases ([DisplayValue](https://github.com/tokio-rs/tracing/blob/bec82286e8b048b031fb070360f397b3da9185d3/tracing-core/src/field.rs#L674) and [DebugValue](https://github.com/tokio-rs/tracing/blob/bec82286e8b048b031fb070360f397b3da9185d3/tracing-core/src/field.rs#L699). Also, I think the event name won't get set to `exception` as it's supposed to; no tracing `message` field => no OpenTelemetry event name, right? There's also logic to copy it to the *span's* attributes (`OpenTelemetryLayer::with_exception_field_propagation` / `ExceptionFieldConfig::propagate`), but that's not the right thing to do by my reading of the semantic conventions above...

...and then I'm not sure `opentelemetry` crates would do the right thing from there anyway. `opentelemetry_datadog` at least doesn't seem to have a conversion from the OpenTelemetry protocol to the Datadog protocol.

I don't know how to accomplish this, but I wish some group of folks could get together and make this integration work end to end. I can file a separate issue in the `opentelemetry` repo about the paragraph above, but I feel like there still may be a gap between the two that I don't know how to fill.

My first impression is that going through these two conversions just seems really janky. It looks like the `opentelemetry` crate [has](https://docs.rs/opentelemetry/0.19.0/opentelemetry/trace/trait.Span.html#method.record_error) this `Span{,Ref}::record_error` method that could do the right thing for the given backend (otlp, datadog, etc). But `tracing_opentelemetry` isn't using it (instead trying to directly follow the opentelemetry semantic conventions), and I think it's not actually using backend-specific logic so I don't see any reason to believe it'd work correctly with `opentelemetry_datadog` anyway.

(Bit of a tangent, but the tracing -> opentelemetry -> datadog conversion seems janky in other ways too. E.g. `tracing` has static/low-cardinality span names. `tracing_opentelemetry` copies those to the OpenTelemetry span names (unless overridden via `otel.name` attribute). But then `opentelemetry_datadog` says in the [crate-level doc comment](https://docs.rs/opentelemetry-datadog/0.7.0/opentelemetry_datadog/) that because of cardinality concerns "This exporter therefore takes a different approach of naming the span with the name of the tracing provider, and using the span name to set the resource_name". So even though my code specifies low-cardinality span names, by default they all get replaced with the useless `opentelemetry-datadog`. It's possible to override this, but it's just a stupid papercut from going through two conversions.)

software-artificer:
Make sure we add `exception.type` field to span events any time we report an error as an exception.

## Motivation
This change is addressing the issue described in #242, where traces that are sent to Datadog that contained errors would not properly link and render, causing an error in Datadog's UI: `Missing error message and stack trace`.

## Solution
Datadog apparently requires at least the `exception.message` and `exception.type` fields to be present.

Closes #242
Closes tokio-rs/tracing#2648

### DIFF
--- a/src/layer/filtered.rs
+++ b/src/layer/filtered.rs
@@ -144,8 +144,10 @@ where
                 OtelDataState::Builder {
                     builder,
                     parent_cx: _,
+                    status,
                 } => {
                     builder.attributes.get_or_insert(Vec::new()).push(key_value);
+                    builder.status = status.clone();
                 }
                 OtelDataState::Context { current_cx } => {
                     let span = current_cx.span();
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -35,6 +35,7 @@ const SPAN_EVENT_COUNT_FIELD: &str = "otel.tracing_event_count";
 const EVENT_EXCEPTION_NAME: &str = "exception";
 const FIELD_EXCEPTION_MESSAGE: &str = "exception.message";
 const FIELD_EXCEPTION_STACKTRACE: &str = "exception.stacktrace";
+const FIELD_EXCEPTION_TYPE: &str = "exception.type";
 
 /// An [OpenTelemetry] propagation layer for use in a project that uses
 /// [tracing].
@@ -287,6 +288,9 @@ impl field::Visit for SpanEventVisitor<'_, '_> {
                     self.event_builder
                         .attributes
                         .push(KeyValue::new(FIELD_EXCEPTION_MESSAGE, format!("{value:?}")));
+                    self.event_builder
+                        .attributes
+                        .push(KeyValue::new(FIELD_EXCEPTION_TYPE, "&str"));
                 } else {
                     self.event_builder
                         .attributes
@@ -326,6 +330,9 @@ impl field::Visit for SpanEventVisitor<'_, '_> {
                     self.event_builder
                         .attributes
                         .push(KeyValue::new(FIELD_EXCEPTION_MESSAGE, format!("{value:?}")));
+                    self.event_builder
+                        .attributes
+                        .push(KeyValue::new(FIELD_EXCEPTION_TYPE, "&dyn fmt::Debug"));
                 } else {
                     self.event_builder
                         .attributes
@@ -367,6 +374,10 @@ impl field::Visit for SpanEventVisitor<'_, '_> {
                 Key::new(FIELD_EXCEPTION_MESSAGE),
                 Value::String(StringValue::from(error_msg.clone())),
             ));
+            self.event_builder.attributes.push(KeyValue::new(
+                Key::new(FIELD_EXCEPTION_TYPE),
+                "&dyn std::error::Error",
+            ));
 
             // NOTE: This is actually not the stacktrace of the exception. This is
             // the "source chain". It represents the heirarchy of errors from the
@@ -390,6 +401,10 @@ impl field::Visit for SpanEventVisitor<'_, '_> {
             attributes.push(KeyValue::new(
                 FIELD_EXCEPTION_MESSAGE,
                 Value::String(error_msg.clone().into()),
+            ));
+            attributes.push(KeyValue::new(
+                FIELD_EXCEPTION_TYPE,
+                "&dyn std::error::Error",
             ));
 
             // NOTE: This is actually not the stacktrace of the exception. This is
@@ -547,6 +562,10 @@ impl field::Visit for SpanAttributeVisitor<'_> {
             self.record(KeyValue::new(
                 Key::new(FIELD_EXCEPTION_MESSAGE),
                 Value::from(error_msg.clone()),
+            ));
+            self.record(KeyValue::new(
+                Key::new(FIELD_EXCEPTION_TYPE),
+                "&dyn std::error::Error",
             ));
 
             // NOTE: This is actually not the stacktrace of the exception. This is
@@ -1738,6 +1757,10 @@ mod tests {
 
         assert_eq!(attributes[FIELD_EXCEPTION_MESSAGE].as_str(), "user error");
         assert_eq!(
+            attributes[FIELD_EXCEPTION_TYPE].as_str(),
+            "&dyn std::error::Error"
+        );
+        assert_eq!(
             attributes[FIELD_EXCEPTION_STACKTRACE],
             Value::Array(
                 vec![
@@ -1881,6 +1904,10 @@ mod tests {
         );
 
         assert_eq!(attributes[FIELD_EXCEPTION_MESSAGE].as_str(), "user error");
+        assert_eq!(
+            attributes[FIELD_EXCEPTION_TYPE].as_str(),
+            "&dyn std::error::Error"
+        );
         assert_eq!(
             attributes[FIELD_EXCEPTION_STACKTRACE],
             Value::Array(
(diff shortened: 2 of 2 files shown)


src/layer/filtered.rs
```diff
@@ -144,8 +144,10 @@ where
                 OtelDataState::Builder {
                     builder,
                     parent_cx: _,
+                    status,
                 } => {
                     builder.attributes.get_or_insert(Vec::new()).push(key_value);
+                    builder.status = status.clone();
```
software-artificer, line 150: Please double-check this for me. At the moment, the default branch if failing to build, because the status field isn't properly matched. I think I am doing the right thing here, but if I should rather ignore the status instead of setting it on the builder, please let me know and I will make an update.

cijothomas (changes requested):
I don't think this should be fixed in this layer. If a particular attribute is required for a backend, then this can be solved by adding the missing pieces elsewhere- like OTel SDK or OTel Collector.

mladedav:
Can the otel SDK conditionally add information? As in if there is `exception.message` in an event, set `exception.type` to `xyz`?

cijothomas:
> Can the otel SDK conditionally add information? As in if there is `exception.message` in an event, set `exception.type` to `xyz`?

Not by default, unless it is required by the specification. But users can add things themselves.

software-artificer:
Just a bit of an update and some extra context. I had a bit of a chat in Discord and discussed how we can do this better. Ideally we would be able to utilize the Valuable crate in the tracing ecosystem to be able to preserve the type information all the way through. This will require some changes in the tracing/core itself and I am exploring how to make this happen. No significant progress thus far because I got some priority changes, but I will come back to this end of the week / start of next week and continue the exploration.

software-artificer:
👋🏻 happy New Year everyone. Just another quick update: I've been tinkering with `Valuable` a little and discovered a gap in `#[instrument]` macro, where there was no way to record neither the return value nor the error value using the `Valuable` trait, so I am addressing this in tokio-rs/tracing#3451. I am also working my way through proper `Valuable` support in `tracing-opentelemetry` that will take advantage of all of this.

Issue #183, opened 2024-11-07
Follows from span not tracked in multithreaded runtime.


mTsBucy1:
### Version
`tracing-opentelemetry: 0.25.0 / 0.27.0` (elaborated below),
`tracing: 0.1.40`

### Platform

`Linux mtstestclient03 6.8.0-40-generic #40~22.04.3-Ubuntu SMP PREEMPT_DYNAMIC Tue Jul 30 17:30:19 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`

### Description

I have this scenario in my code:
1. I spawn (multiple) task, which loops and call async instrumented function.
2. this function gets its `tracing::Span::current().id()` and sends it through a tokio mpsc channel.
3. Another task reads this, spawns a instrumented async function, which follows from the received id.

Suddenly my code paniced on startup with 
```
thread 'tokio-runtime-worker' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-opentelemetry-0.25.0/src/layer.rs:953:18:
Missing otel data span extensions
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-subscriber-0.3.18/src/registry/sharded.rs:437:58:
Mutex poisoned: PoisonError { .. }
[stack trace...]
panic in a destructor during cleanup
thread caused non-unwinding panic. aborting.
```
This happened here in the code https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.25.0/src/layer.rs#L953.
So far, this hasn't happened again.

I wrote a test, now on the current branch (0.27.0), to try to replicate this behaviour, but it also didn't show the same panic. However, I noticed another odd behaviour: The asserted count fails for a `tokio new_multi_thread` runtime, but succeeds with the `tokio new_current_thread` one. I am not sure whether these are indeed related, but it seems to be a bug in its own right.

```rust
// in tests/follows_from.rs

#[test]
fn trace_follows_from_diff_thread() {
    use tokio::sync::mpsc;

    #[tracing::instrument]
    async fn span_follows_this(chan: &mpsc::Sender<Option<tracing::Id>>) {
        chan.send(tracing::Span::current().id())
            .await
            .expect("Channel should be open");
    }

    #[tracing::instrument(follows_from = [cause])]
[...]

PR #244, opened 2025-12-08
Fix the panic in `follows_from`

mTsBucy1:
## Motivation

I have been experiencing crashes in my program approximately every 10 minutes due to <https://github.com/tokio-rs/tracing-opentelemetry/issues/183>. I have now investigated and found the problem, reproduced it with a test and fixed the panic. Additionally I was missing some documentation on this behaviour.

## Solution

This is fixed by instead of `expect`-ing the extension data to instead use `let-else` to match and ignore if it is not present. This behaviour also exists in other implementations in the same file.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -65,6 +65,15 @@ where
 
 /// Construct a layer to track spans via [OpenTelemetry].
 ///
+/// It will convert the tracing spans to OpenTelemetry spans
+/// and tracing events which are in a span to events in the currently active OpenTelemetry span.
+/// Child-Parent links will be automatically translated as well.
+/// Follows-From links will also be generated, but only if the originating span has not been closed yet.
+///
+/// Various translations exist between semantic conventions of OpenTelemetry and `tracing`,
+/// such as including level information as fields, or mapping error fields to exceptions.
+/// See the various `layer().with_*` functions for configuration of these features.
+///
 /// [OpenTelemetry]: https://opentelemetry.io
 ///
 /// # Examples


mTsBucy1:
(note that CI fails due to the lastest commit in the main branch, not my branch)

tests/follows_from.rs
```diff
@@ -57,3 +59,51 @@ fn trace_follows_from_closed() {
...
+    /// This layer will take 20ms to process
+    #[derive(Clone)]
+    struct WaitOnCloseLayer;
+    impl<S> Layer<S> for WaitOnCloseLayer
+    where
+        S: Subscriber,
+    {
+        fn on_close(&self, _id: tracing::Id, _ctx: tracing_subscriber::layer::Context<'_, S>) {
+            // Some longer running computation
+            std::thread::sleep(Duration::from_millis(20));
```
mladedav, line 74: [Sleep is not a synchronization primitive](https://xeiaso.net/blog/nosleep/), but I'll merge this as is. However, I don't think the code would be more complicated if you had `WaitOnCloseLayer(Notify)` instead.

mladedav (approved):
Thanks

mladedav:
Sorry for the close, it is merged into the default branch and you're still the commits' author. I tried to merge it locally to keep the history clean since you split and named the commits well but in the process GitHub found that all the changes have been applied and it closed this automatically.

So the code is merged, just the red close status here should be purple.

tests/follows_from.rs
mTsBucy1: yes, thanks for the pointer, I had the same thought in bed after implementing this :P

Issue #250, opened 2026-01-26
deadlock with `opentelemetry` and `tracing`


aome510:
### Version

```
tracing-opentelemetry v0.32.1 (/Users/aome510/Projects/tracing-opentelemetry)
│   └── tracing v0.1.44
│       ├── tracing-attributes v0.1.31 (proc-macro)
│       └── tracing-core v0.1.36
├── tracing v0.1.44 (*)
├── tracing-core v0.1.36 (*)
├── tracing-log v0.2.0
│   └── tracing-core v0.1.36 (*)
└── tracing-subscriber v0.3.22
    └── tracing-core v0.1.36 (*)
│   │       │   │   └── tracing v0.1.44 (*)
│   │       │   └── tracing v0.1.44 (*)
│   │       │   └── tracing v0.1.44 (*)
│   │   │   └── tracing v0.1.44 (*)
│   └── tracing v0.1.44 (*)
├── tracing v0.1.44 (*)
├── tracing-error v0.2.1
│   ├── tracing v0.1.44 (*)
│   └── tracing-subscriber v0.3.22 (*)
└── tracing-subscriber v0.3.22 (*)
```

### Description

I recently encountered a deadlock happened because an interaction between `tracing`, `opentelemetry` and this crate. `opentelemetry` uses `tracing` for their [internal logging](https://github.com/open-telemetry/opentelemetry-rust/blob/1a47446dcc546d4a6f0c1e0e516503e5428aa5e7/opentelemetry/Cargo.toml#L41) (enabled by default). If the logging API is called inside the below block (`opentelemetry` has a few places that do that inside the block)

https://github.com/tokio-rs/tracing-opentelemetry/blob/feedb2276880eb5f5be0884b165ab4707d8d3444/src/layer.rs#L1207-L1210

it will result in a deadlock. Stacktrace when the deadlock happens

```
0x0000723f78fcc23f in clock_nanosleep () from target:/usr/lib/x86_64-linux-gnu/libc.so.6
#0  0x0000723f7900795d in syscall () from target:/usr/lib/x86_64-linux-gnu/libc.so.6
#1  0x000059c6f9599bd9 in std::sys::pal::unix::futex::futex_wait () at library/std/src/sys/pal/unix/futex.rs:72
#2  std::sys::sync::rwlock::futex::RwLock::write_contended () at library/std/src/sys/sync/rwlock/futex.rs:257
[...]

PR #251, opened 2026-01-28
fix: avoid deadlock when entering a span

mladedav:
## Motivation

There is a potential deadlock described in #250 where we lock `ExtensionsMut` when entering a span, we let `opentelemetry` activate the context, it tries to log through `tracing` in that operation which in turn calls `on_event` in this layer and we try to lock `ExtensionsMut` again.

## Solution

Instead of accessing span's internal state through `ExtensionsMut`, we store an `Arc<Mutex<OtelData>>`, clone the `Arc` and then lock it to access it. This allows us to hold onto the state without deadlocking when other layers try to lock the extensions in re-entrant calls.

There's a test to check for this but it needs to be run with `RUSTFLAGS='--cfg reentrant_tracing_test'` environment. I did not put that into the CI because changing the environment forces recompilation of every dependency. So I'm still looking whether I'm able to do this a bit better or if I'll just drop the test.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -1,12 +1,13 @@
 use crate::stack::IdValueStack;
-use crate::{OtelData, OtelDataState};
+use crate::{OtelData, OtelDataLock, OtelDataState};
 pub use filtered::FilteredOpenTelemetryLayer;
 use opentelemetry::ContextGuard;
 use opentelemetry::{
     trace::{self as otel, noop, Span, SpanBuilder, SpanKind, Status, TraceContextExt},
     Context as OtelContext, Key, KeyValue, StringValue, Value,
 };
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
+use std::sync::Arc;
 use std::thread;
 #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"))))]
 use std::time::Instant;
@@ -19,7 +20,7 @@ use tracing_core::{field, Event, Subscriber};
 use tracing_log::NormalizeEvent;
 use tracing_subscriber::layer::Context;
 use tracing_subscriber::layer::Filter;
-use tracing_subscriber::registry::{ExtensionsMut, LookupSpan};
+use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::Layer;
 #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
 use web_time::Instant;
@@ -35,6 +36,10 @@ const SPAN_EVENT_COUNT_FIELD: &str = "otel.tracing_event_count";
 const EVENT_EXCEPTION_NAME: &str = "exception";
 const FIELD_EXCEPTION_MESSAGE: &str = "exception.message";
 const FIELD_EXCEPTION_STACKTRACE: &str = "exception.stacktrace";
+
+std::thread_local! {
+    static INSIDE_TRACING: Cell<bool> = const { Cell::new(false) };
+}
 
 /// An [OpenTelemetry] propagation layer for use in a project that uses
 /// [tracing].
@@ -125,7 +130,7 @@ pub(crate) struct WithContext {
     ///
     #[allow(clippy::type_complexity)]
     pub(crate) with_activated_otel_context:
-        fn(&tracing::Dispatch, &mut ExtensionsMut<'_>, f: &mut dyn FnMut(&OtelContext)),
+        fn(&tracing::Dispatch, &span::Id, f: &mut dyn FnMut(&OtelContext)),
 }
 
 impl WithContext {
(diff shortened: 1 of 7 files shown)


mladedav:
I'll have to look closer at what's happening with the `RefCell` when setting the dispatch. It's at least panicking consistently.

mladedav:
Changing the `Dispatch` will not work. If we're not using the global dispatcher, the ref cell is already borrowed because someone is logging something so when we try to override it from within these methods, we're trying to mutably borrow a ref cell that's already borrowed, leading to a panic.

Now I'm trying to figure out if I can use `event_enabled` to filter out events inside the critical sections but that could still leave us open to tracing span usage inside opentelemetry or one of its dependencies.

Maybe we'll have to just rework this to not hold onto the locks while calling out to opentelemetry.

mladedav:
@bantonsson if  you have some time, would you mind taking a look? This reworks the way you added the method to get to OpenTelemetry context from `SpanRef` so I just want to make sure I'm not missing something.

The solution right now has pretty bad performance regression (one benchmark went up by 23 %) and I hope to get that a bit under control, but in the worst case, bad performance sounds better than a deadlock.

bantonsson:
Hey @mladedav. I completely missed this notification. I'll try to look at it today or tomorrow.

src/layer.rs
```diff
@@ -36,6 +38,10 @@ const EVENT_EXCEPTION_NAME: &str = "exception";
 const FIELD_EXCEPTION_MESSAGE: &str = "exception.message";
 const FIELD_EXCEPTION_STACKTRACE: &str = "exception.stacktrace";
 
+std::thread_local! {
+    static INSIDE_TRACING: AtomicBool = const { AtomicBool::new(false) }
+}
```
bantonsson, line 43: Since this is a thread local, it will only be accessed by the current thread, so it doesn't need atomic ordering.

```suggestion
std::thread_local! {
    static INSIDE_TRACING: Cell<bool> = const { Cell::new(false) }
}
```

bantonsson (commented):
Finally got around to this. I'll continue to look at it, but one thing immediately caught my eye.

src/layer.rs
mladedav: Thanks, changed to `Cell`.

houseme (approved):

houseme (changes requested):
I think this is very close, but I don't think it fully fixes #250 yet.

The main problem is that `on_enter` still holds the per-span `OtelData` mutex across `cx.clone().attach()`. Issue #250 shows that `Context::attach()` can emit tracing logs, which re-enter `on_event`. In this patch, `on_event` then looks up the same span and tries to lock the same `OtelDataLock` again. So the original deadlock through `ExtensionsMut` seems to have turned into a deadlock on the per-span `Mutex`.

`INSIDE_TRACING` only protects `start_cx()`, but it is cleared before `attach()` runs, so it does not cover the path reported in #250.

I also don't think the cfg-gated regression check proves the bug is fixed yet, because it injects re-entrant tracing inside `start_with_context`, while the reported stack in #250 goes through `Context::attach()` from `on_enter`.

Because of that, my review is request-changes: I think we need to avoid holding the span-state lock across `attach()`, or otherwise suppress re-entrant tracing for that whole critical section, and add a regression test that exercises the `attach()` path directly.

mladedav:
I think you're right though I have trouble triggering it without the `cfg` and strictly speaking, we should do this for _every_ foreign library call.

It would be great to have more robust solution but for now, I just added another location doing the same thing. Which will break the moment someone accidentally nests these.

Or we could just set the flag the moment we enter any of the layer method I just worry it'll be unnecessarily expensive. I hope I get the time to revisit this someday.

mladedav:
I changed it once more and added one more place.

This should definitely help with the reported issue though there might be other places from where `opentelemetry` logs and we might have to play a bit of whack-a-mole or just bite the bullet and set  the cell on each enter into any layer function and check it in every function. But that needs to be benchmarked, until now, this might hurt less and should help.

src/lib.rs
```diff
@@ -122,61 +122,46 @@ mod span_ext;
...
-impl OtelData {
-    /// Gets the trace ID of the span.
-    ///
-    /// Returns `None` if the context has not been built yet. This can be forced e.g. by calling
-    /// [`context`] on the span (not on `OtelData`) or if [context activation] was not explicitly
-    /// opted-out of, simply entering the span for the first time.
-    ///
-    /// [`context`]: OpenTelemetrySpanExt::context
-    /// [context activation]: OpenTelemetryLayer::with_context_activation
-    pub fn trace_id(&self) -> Option<opentelemetry::TraceId> {
```
lake-armitage-anaplan, line 153: Does this API have any analogue now? We're using the below snippet to grab the current trace id for logging and I can't adopt the fixed version.

mladedav: Yes, see [this example](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/otel_context.rs).

The most relevant parts are [this](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/otel_context.rs#L86-L95) and [that](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/otel_context.rs#L71).

lake-armitage-anaplan: thanks for the pointer, managed to get this figured out :)

PR #252, opened 2026-01-28
Mimimize package size

weiznich:
## Motivation
During a dependency review we noticed that tracing-opentelemetry includes a binary file (picture) in the source code uploaded to crates.io. While this is on it's own not a problem, its still wasteful as it copies this picture to each system using tracing-opentelemetry as dependency. Additionally it includes binary and therefore harder to review data in the published code. This PR changes the link in the readme to rely on a copy of the image hosted by github instead. 

## Solution

This PR sets a `include` configuration to explicitly list which files should be part of the published package.

This reduces the package size from 41 files, 431.4KiB (153.4KiB compressed) to 24 files, 298.3KiB (61.4KiB compressed), which directly translates to 500GB less traffic per month for crates.io

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ keywords = ["tracing", "opentelemetry", "jaeger", "zipkin", "async"]
 license = "MIT"
 edition = "2021"
 rust-version = "1.75.0"
+include = ["src/**/*.rs", "tests/**/*.rs", "CHANGELOG.md", "README.md", "Cargo.toml", "LICENSE"]
 
 [features]
 default = ["tracing-log", "metrics"]
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ $ cargo run --example opentelemetry-otlp
 $ firefox http://localhost:16686/
 ```
 
-![Jaeger UI](trace.png)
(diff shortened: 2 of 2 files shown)


mladedav:
Thanks! May I ask why tests are included and examples and benches are not? I would expect we'd want just `src` or all four.

I thought I've also seen somewhere some talk about having examples shown on docs.rs but I could find only [this old issue](https://github.com/rust-lang/docs.rs/issues/132) so maybe I hallucinated that one.

weiznich:
I included tests and excluded examples and benches because various other maintainers expressed such a preference in the past. I personally would rather also remove the tests because they are unused by a large majority of users as well. Notable exceptions there are crater runs, people that vendor tracing-opentelemetry and then execute the tests there and linux distribution maintainers (which might execute the tests during their build processes).


> I thought I've also seen somewhere some talk about having examples shown on docs.rs but I could find only https://github.com/rust-lang/docs.rs/issues/132 so maybe I hallucinated that one.

That's likely this unstable feature: https://doc.rust-lang.org/rustdoc/scraped-examples.html

I did remove the examples from the published package because it was not enabled.

mladedav:
Thanks for the explanation, that makes sense.

PR #256, opened 2026-03-13
feat: support updating span name (#240)

### LINKED ISSUE #240: Unable to update `otel.name` field
### Version

```
tracing v0.1.41
tracing-attributes v0.1.30
tracing-core v0.1.34
tracing-opentelemetry v0.32.0
tracing-log v0.2.0
tracing-subscriber v0.3.20
tracing-serde v0.2.0
```

### Platform

```
Darwin grqmuxsirm.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:29:54 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T8122 arm64
```

### Description

I'm not able to update the special `otel.name` attribute after it's set on a span. For context, I'm using `tower_http::trace::TraceLayer` on my root axum router with a custom `MakeSpan` to set the `otel.name` to include the matched axum route. However, I also have some nested routers (added with [Router::nest_service](https://docs.rs/axum/latest/axum/struct.Router.html#method.nest_service)). Because the routers are nested, the matched path does not include the full path from the nested service. So, I was trying to attach an additional layer to the nested service to update the `otel.name` value. This seems to partially work -- traces appear to have the correct value when I view them with the log subscriber. However, the root span name is not correct in the otel collector (grafana in my case). I think it's expected that the span field is not updated if it already has a value, but I'm seeing the same behavior even if I set the value to `tracing::field::Empty` initially.

I tried a few different approaches to update the value:

1. Using `Span::record`
```rust
let span = tracing::info_span!("A", otel.name = "otel-A", http.route = "/pathA");
span.in_scope(|| {
    Span::current().record("otel.name", "otel-A updated");
    Span::current().record("http.route", "/pathA/updated");
    tracing::info!("/pathA event");
});
```
which results in:
```
2025-11-27T20:40:40.972236Z  INFO A{otel.name="otel-A" http.route="/pathA" otel.name="otel-A updated" http.route="/pathA/updated"}: opentelemetry_otlp: /pathA event
```
Note the duplicate `otel.name` and `http.route` fields and the root span does not have the correct name in grafana (screenshot below).

2. Using `Span::record` with `tracing::field::Empty`
```rust
let span = tracing::info_span!("B", otel.name = tracing::field::Empty, http.route = tracing::field::Empty);
span.in_scope(|| {
    Span::current().record("otel.name", "otel-B updated");
    Span::current().record("http.route", "/pathB/updated");
    tracing::info!("/pathB event");
});
```
which results in:
```
2025-11-27T20:40:40.972295Z  INFO B{otel.name="otel-B updated" http.route="/pathB/updated"}: opentelemetry_otlp: /pathB event
```
Note that the fields have the correct values, but the root span does not have the correct name in grafana (screenshot below)

3. Using `OpenTelemetrySpanExt::set_attribute`:
```rust
let span = tracing::info_span!("C", otel.name = "otel-C", http.route = "/pathC");
span.in_scope(|| {
    Span::current().set_attribute("otel.name", "otel-C updated");
    Span::current().set_attribute("http.route", "/pathC/updated");
    tracing::info!("/pathC event");
});
```
which results in:
```
2025-11-27T20:40:40.972349Z  INFO C{otel.name="otel-C" http.route="/pathC"}: opentelemetry_otlp: /pathC event
```
Note that the fields are not updated.

4. Using `OpenTelemetrySpanExt::set_attribute` with `tracing::field::Empty`:
```rust
let span = tracing::info_span!("D", otel.name = tracing::field::Empty, http.route = tracing::field::Empty);
span.in_scope(|| {
    Span::current().set_attribute("otel.name", "otel-D updated");
    Span::current().set_attribute("http.route", "/pathD/updated");
    tracing::info!("/pathD event");
});
```
which results in
```
2025-11-27T20:40:40.972435Z  INFO D: opentelemetry_otlp: /pathD event
```
Note that the fields are not present at all.

In all of the above cases, the updated `otel.name` value does not appear to be reported to the OTEL collector (grafana in my case) even if it appears to be correct in the log output.

Image

However, the `http.route` is correct when it's originally set to `tracing::field::Empty`, as in examples 2 and 4 (B and D) (located next to the timestamp in the below screenshots):

A

Image

B

Image

C

Image

D

Image

tommilligan:
Closes #240 

This PR exposes the underlying `update_name` method from `opentelemetry`, allowing the span name to be updated after creation.

This follows the established pattern for `set_attribute` and `set_status`.

## Motivation

I hit the same issue as described [here](https://github.com/tokio-rs/tracing-opentelemetry/issues/240#issuecomment-3587483084), which is not being able to set a name on a span after creation.

## Solution

Implemented standard method following existing code pattern, basically just forwarding to the underlying library.

### DIFF
--- a/src/span_ext.rs
+++ b/src/span_ext.rs
@@ -170,6 +170,29 @@ pub trait OpenTelemetrySpanExt {
     /// ```            
     fn set_status(&self, status: Status);
 
+    /// Sets an OpenTelemetry name for this span.
+    /// This is useful for setting the name of a span that has a dynamic name, where the name is not
+    /// known when starting the span.
+    ///
+    /// For spans where the name is dynamic but known at creation, use the `otel.name` attribute
+    /// with the `span!` macro.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use tracing_opentelemetry::OpenTelemetrySpanExt;
+    /// use tracing::Span;
+    ///
+    /// /// // Generate a tracing span as usual
+    /// let app_root = tracing::span!(tracing::Level::INFO, "app_start");
+    ///
+    /// // Set the name of the span to `app_start (42)`.
+    /// let dynamic_data = 42;
+    /// let name = format!("app_start ({})", dynamic_data);
+    /// app_root.update_name(name.into());
+    /// ```
+    fn update_name(&self, name: Cow<'static, str>);
+
     /// Adds an OpenTelemetry event directly to this span, bypassing `tracing::event!`.
     /// This allows for adding events with dynamic attribute keys, similar to `set_attribute` for span attributes.
     /// Events are added with the current timestamp.
@@ -396,6 +419,24 @@ impl OpenTelemetrySpanExt for tracing::Span {
                 OtelDataState::Context { current_cx } => {
                     let span = current_cx.span();
                     span.set_status(status.take().unwrap());
+                }
+            });
+        });
+    }
+
+    fn update_name(&self, name: Cow<'static, str>) {
+        self.with_subscriber(move |(id, subscriber)| {
+            let mut name = Some(name);
+            let Some(get_context) = subscriber.downcast_ref::<WithContext>() else {
+                return;
+            };
+            get_context.with_context(subscriber, id, move |data| match &mut data.state {
+                OtelDataState::Builder { builder, .. } => {
+                    builder.name = name.take().unwrap();
+                }
+                OtelDataState::Context { current_cx } => {
+                    let span = current_cx.span();
+                    span.update_name(name.take().unwrap());
                 }
             });
         });


tommilligan:
Tagging @mladedav for review as you were part of the original issue thread.

mladedav:
Why not the proposed solution there of `tracing::Span::current().context().span().update_name("name");`? I'd rather not expose individually every feature of opentelemetry and have people use the context directly if they need to.

I'd rather deprecate and eventually remove the attribute and status methods to be completely honest.

tommilligan:
I've tried that workaround but not had any success (name of the span is not updated when it is emitted). I don't quite understand why this is in my specific setup - having the unit tests here passing gives me more confidence than hacking around with extension traits in my codebase.

I will try backporting this fix to the version I'm actually using and seeing if it fixes my issue in situ - that's a fair point to try.

I'm currently working around this with a `transform` at the Opentelemetry Collector level that just copies `rpc.method` to the span name when set, as that's what I'm trying to achieve.

I'll also try and read a bit more about the differences between `tracing` and `tracing-opentelemetry` to understand how they interact here.

mladedav:
To save you just a bit of time, `tracing` defines a `Layer<S>` trait which crates like `tracing-opentelemetry` implement. Then when you do `tracing::info!(...)`, tracing calls `on_event` on our `Layer<S>` implementation and the layer is then free to do anything it wants with that.

What might also be importatnt are slight differences between `tracing::Span` and opentelemetry spans, for example the former needs a static name at creation time and it cannot be changed later while opentelemetry allows it to change and the name can also be runtime-generated.

tommilligan:
Thanks for the context; but it looks like something much simpler. I rewrote the my unit test cases onto `da83e89` (v0.31.0 tracing-opentelemetry, released 2025-06-02) and the suggested fix no longer works (see branch [here](https://github.com/tokio-rs/tracing-opentelemetry/compare/v0.1.x...tommilligan:tracing-opentelemetry:feat-update-span-name-0.31) - `cargo test --all-features` fails).

I think that means I'll fork 0.31 and apply my fix for now, and then move to the suggested extension trait approach once we upgrade.

Thanks for your help!

tommilligan:
Closed as already supported via https://github.com/tokio-rs/tracing-opentelemetry/issues/240#issuecomment-3587457299 in latest version.

PR #260, opened 2026-05-16
fix: avoid reentrant deadlocks when activating otel context

### LINKED ISSUE #250: deadlock with `opentelemetry` and `tracing`

houseme:
## Summary

Fixes #250. Follow-up to #251.

This PR removes reentrant deadlock windows in context activation by ensuring OpenTelemetry API calls do not run while holding the span-local `OtelData` lock.

Key changes:
- add transient `OtelDataState::Starting` and coordinate startup with `Condvar`
- defer context-path OTel calls (`add_event`, `add_link`, `set_attribute`, `set_status`) until after lock release
- keep `on_close`/`follows_from` paths coordinated with in-flight startup
- add a regression test for the reentrant `attach()` path

## Review Follow-ups

- restore focused docs for context lookup helpers
- simplify lookup shape to `Option<OtelDataLock>`
- add `ReentrantTracingGuard::with(...)` and migrate guarded call-sites

## Validation

- `cargo fmt`
- `cargo test --tests --all-features`
- `RUSTFLAGS='--cfg __reentrant_tracing_test' cargo test --test otel_context`

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -1008,7 +1008,6 @@ where
         let span = subscriber
             .span(id)
             .expect("registry should have a span for the current ID");
-
         let lookup = span.extensions().get::<OtelDataLock>().cloned();
         lookup
     }
@@ -1083,8 +1082,7 @@ where
             .downcast_ref::<OpenTelemetryLayer<S, T>>()
             .expect("layer should downcast to expected type; this is a bug!");
         if let Some(otel_data) = Self::lookup_otel_data(dispatch, id) {
-            let current_cx = layer.ensure_context_snapshot(&otel_data);
-            f(&current_cx);
+            f(&layer.ensure_context_snapshot(&otel_data));
         }
     }
 
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -54,6 +54,11 @@ impl ReentrantTracingGuard {
         });
         Self { previous }
     }
+
+    fn with<T>(f: impl FnOnce() -> T) -> T {
+        let _guard = Self::enter();
+        f()
+    }
 }
 
 impl Drop for ReentrantTracingGuard {
@@ -212,11 +217,6 @@ impl WithContext {
     ) {
         (self.with_activated_otel_context)(dispatch, id, &mut f)
     }
-}
-
-enum SpanLookup {
-    Missing,
-    Found(OtelDataLock),
 }
 
 fn str_to_span_kind(s: &str) -> Option<otel::SpanKind> {
@@ -998,7 +998,8 @@ where
         }
     }
 
-    fn lookup_otel_data(dispatch: &tracing::Dispatch, id: &span::Id) -> SpanLookup {
+    /// Returns the span-local OpenTelemetry storage for the given tracing span ID.
+    fn lookup_otel_data(dispatch: &tracing::Dispatch, id: &span::Id) -> Option<OtelDataLock> {
         let subscriber = dispatch
             .downcast_ref::<S>()
             .expect("subscriber should downcast to expected type; this is a bug!");
@@ -1006,23 +1007,24 @@ where
             .span(id)
             .expect("registry should have a span for the current ID");
 
-        let lookup = span
-            .extensions()
-            .get::<OtelDataLock>()
-            .cloned()
-            .map(SpanLookup::Found)
-            .unwrap_or(SpanLookup::Missing);
+        let lookup = span.extensions().get::<OtelDataLock>().cloned();
         lookup
     }
 
+    /// Provides mutable access to a span's OpenTelemetry state.
+    ///
+    /// If span startup is in progress, this waits for startup to finish before
+    /// invoking `f`.
     fn get_context(dispatch: &tracing::Dispatch, id: &span::Id, f: &mut dyn FnMut(&mut OtelData)) {
-        if let SpanLookup::Found(otel_data) = Self::lookup_otel_data(dispatch, id) {
+        if let Some(otel_data) = Self::lookup_otel_data(dispatch, id) {
             let mut locked = otel_data.lock();
             locked = otel_data.wait_while_starting(locked);
             f(&mut locked);
         }
     }
 
+    /// Activates a span's OpenTelemetry context and then provides mutable access
+    /// to the span state.
     fn get_activated_context(
         dispatch: &tracing::Dispatch,
         id: &span::Id,
@@ -1032,7 +1034,7 @@ where
             .downcast_ref::<OpenTelemetryLayer<S, T>>()
             .expect("layer should downcast to expected type; this is a bug!");
 
-        if let SpanLookup::Found(otel_data) = Self::lookup_otel_data(dispatch, id) {
+        if let Some(otel_data) = Self::lookup_otel_data(dispatch, id) {
             let current_cx = layer.ensure_context_snapshot(&otel_data);
             let mut locked = otel_data.lock();
             locked = otel_data.wait_while_starting(locked);
@@ -1137,14 +1139,13 @@ where
                 parent_cx,
                 status,
             } => {
-                let current_cx = {
-                    let _guard = ReentrantTracingGuard::enter();
+                let current_cx = ReentrantTracingGuard::with(|| {
                     #[cfg(__reentrant_tracing_test)]
                     tracing::info!("This should not deadlock...");
                     let mut span = builder.start_with_context(&self.tracer, &parent_cx);
                     span.set_status(status);
                     parent_cx.with_span(span)
-                };
+                });
 
                 let mut locked = otel_data.lock();
                 locked.state = OtelDataState::Context {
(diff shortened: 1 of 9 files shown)


tests/otel_context.rs
```diff
@@ -175,3 +175,30 @@ fn test_span_ref_ext_from_separate_layer() {
...
+    assert_eq!(
+        extracted_contexts.len(),
+        1,
+        "Expected the custom layer to extract the reentrant parent context"
+    );
+
+    let extracted = extracted_contexts[0].span().span_context().span_id();
+    let exported = spans[0].span_context.span_id();
+    assert_eq!(extracted, exported, "Reentrant attach should still yield the active span context");
+}
```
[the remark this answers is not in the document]
houseme: Adopted in commit 873b8ef. I changed the instrumentation cfg from #[cfg(all(test, __reentrant_tracing_test))] to #[cfg(__reentrant_tracing_test)] so integration-test builds include the injected reentrant path. Re-run: RUSTFLAGS=--cfg __reentrant_tracing_test cargo test --test otel_context, now observing the injected This should not deadlock logs and passing.

src/layer.rs
```diff
@@ -1311,25 +1380,34 @@ where
...
-            let follows_context =
-                self.with_started_cx(&mut follows_locked, &|cx| cx.span().span_context().clone());
-            match &mut data.lock().state {
-                OtelDataState::Builder { builder, .. } => {
-                    if let Some(ref mut links) = builder.links {
-                        links.push(otel::Link::with_context(follows_context));
-                    } else {
-                        builder.links = Some(vec![otel::Link::with_context(follows_context)]);
+            let follows_context = self
+                .ensure_context_snapshot(&follows_data)
```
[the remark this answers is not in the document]
houseme: Adopted in commit 873b8ef. In on_follows_from we now obtain the follows context snapshot before dropping follows_extensions, then release extensions after snapshot extraction. This narrows the close-race window called out in this thread.

src/layer.rs
```diff
@@ -955,44 +998,31 @@ where
...
+            .map(SpanLookup::Found)
+            .unwrap_or(SpanLookup::Missing);
+        lookup
+    }
+
+    fn get_context(dispatch: &tracing::Dispatch, id: &span::Id, f: &mut dyn FnMut(&mut OtelData)) {
+        if let SpanLookup::Found(otel_data) = Self::lookup_otel_data(dispatch, id) {
+            let mut locked = otel_data.lock();
+            locked = otel_data.wait_while_starting(locked);
+            f(&mut locked);
```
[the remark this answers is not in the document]
houseme: Adopted in commit 873b8ef. I moved OpenTelemetrySpanExt context-path OTel calls out of the OtelData lock via deferred actions (add_link/set_attribute/set_status/add_event), so reentrant tracing no longer reacquires the same lock in this path. Revalidated with clippy -D warnings and the reentrant otel_context test.

mladedav:
You don't need to tag me in every message, I gett a notification either way.

It's also fine if you use AI in your workflow, but in such cases you should make the content consumable by normal people. The description is extremely verbose describing way too much detail about things we've basically just talked about elsewhere.

Ideally write that part yourself if possible, it shows you actually understand the proposed changes as well as giving me the feeling that there's a human being on the other side, not just a model. But yeah, just make it readable is the bottom line.

src/layer.rs
```diff
@@ -955,44 +998,31 @@ where
...
-    ///
-    /// This function retrieves the span from the subscriber's registry using the provided
-    /// span ID, activates the OTel `Context` in the span's `OtelData` if present, and then
-    /// applies the callback function `f` to the `OtelData`.
-    ///
-    /// # Parameters
-    ///
-    /// * `dispatch` - The tracing dispatch to downcast and retrieve the span from
-    /// * `id` - The span ID to look up in the registry
-    /// * `f` - The closure to invoke with mutable access to the span's `OtelData`
```
mladedav, line 1004: Why were docs deleted?

src/layer.rs
```diff
@@ -176,6 +214,11 @@ impl WithContext {
     }
 }
 
+enum SpanLookup {
```
mladedav: This is built exactly once in this diff with having an `Option` and calling basically `.map_or(Self::Missing, Self::Found)` on it. How is this better than `Option<OtelDataLock>`?

src/layer.rs
```diff
@@ -41,6 +41,44 @@ std::thread_local! {
     static INSIDE_TRACING: Cell<bool> = const { Cell::new(false) };
 }
 
+struct ReentrantTracingGuard {
+    previous: bool,
+}
+
+impl ReentrantTracingGuard {
+    fn enter() -> Self {
```
mladedav, line 49: In a perfect world, you want something like `Guard::with(fn: FnOnce() -> T)` rather than these guards because that way you prevent people from dropping out of order.

mladedav (commented):
Out of curiosity, is this AI-assisted or AI-written?

src/layer.rs
houseme: Agreed. I added ReentrantTracingGuard::with(...) and migrated guarded call-sites to the closure form so guard lifetime cannot be dropped out of order.

src/layer.rs
houseme: Agreed. I simplified this to Option<OtelDataLock>; SpanLookup was unnecessary indirection in this path.

src/layer.rs
houseme: Incorrectly deleted, already recovered.

PR #261, opened 2026-05-16
chore(ci): upgrade workflow action dependencies and add cargo-deny baseline config

houseme:
## Summary

This PR updates CI workflow action dependencies and introduces a baseline `cargo-deny` configuration.

### Workflow updates

- Upgrade all `actions/checkout` references from `@v4` to `@v6`:
  - `.github/workflows/CI.yml`
  - `.github/workflows/audit.yml`
  - `.github/workflows/release.yml`

- Keep security audit action on:
  - `rustsec/audit-check@v2`

### New config

- Add `deny.toml` with an initial policy:
  - enable `all-features` for dependency graph checks
  - deny yanked crates
  - set a curated license allowlist
  - warn on multiple versions, deny wildcard deps
  - restrict registries/git sources to known allowed origins

## Why

- Keep GitHub Actions dependencies up to date
- Improve supply-chain and license governance readiness with a repository-level `cargo-deny` baseline

## Scope

- CI/workflow and dependency policy config only
- No runtime code changes

cc @mladedav

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -35,7 +35,7 @@ jobs:
     name: cargo check
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@stable
     - name: Check
       run: cargo check --all --tests --benches
@@ -46,7 +46,7 @@ jobs:
     needs: check
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@stable
       with:
         components: rustfmt
@@ -58,7 +58,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: check
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@stable
       with:
         components: clippy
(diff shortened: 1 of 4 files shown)


mladedav:
We can allow yanked crates assuming we update dependencies in the CI job. We don't want to update them for build jobs. As long as cargo can resolve a version we don't care.

houseme:
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

mladedav:
Yeah, I'm find with updating the actions versions.

I just don't think CI here should be broken by a yanked version that happens to be in our lock file. Which we can work around by updating dependencies in ci before calling cargo deny and letting cargo find the highest working version.

So from the deny.toml, the warning about yanked dependencies should be gone.

houseme:
Thanks for the feedback, addressed in this update.

Changes made:
- Relaxed yanked handling in `deny.toml`:
  - `advisories.yanked = "allow"`
- Removed the license check section from `deny.toml` for now.

`cargo deny` stays in the audit workflow, but with policy aligned to avoid failing CI on yanked versions in the lockfile.

I can revisit the license policy in a separate PR/discussion.

Issue #217, opened 2025-06-24
`otel.name` attribute should apply to events as well as spans


shaylew:
### Motivation

Currently there's a special/reserved attribute `otel.name` that can be used to send a dynamic value as the name of a span, which is supported just fine in opentelemetry but not otherwise a thing in the broader `tracing` environment.

This only works for spans, not for events, which is a little surprising. The use cases for dynamic event names are more or less the same as for dynamic span names, and I imagine the implementation would look similar too. 

### Proposal

The `otel.name` field, when used on a span, should set the name of the resulting span (overriding whatever static name is present). The other `otel.xxx` attributes don't make any sense on events and should continue not to apply. 

Docs should probably mention for completeness that `otel.name` works on events as well as spans, though maybe that's less of a priority -- it's easy to assume it works the same, the surprise is that it currently does not.

### Alternatives

Just making `otel.name` work for both seems like the approach that's most concordant with the existing API. I suppose there could be a separate namespace for span and event magic attributes, but that horse has more or less already flown with `otel.name` being un-namespaced.

I _have_ been asked by coworkers "hey, what on earth is that `otel.name` doing", and perhaps exporting the attribute names which have special meanings as constants would be a chance to attach more discoverable documentation to them and distinguish the span- and event-applicable ones without any breaking change, but that's really orthogonal to the main issue of getting dynamic names on events working some way or other.

mladedav:
> the main issue of getting dynamic names on events working some way or other.

The event name is set to whatever is in `message` which is the magic unnamed field in `tracing`. So if you do `tracing::info!(?foo, %bar, "This is the {}th log.", 5);`, then the name of the event will be set to `This is the 5th log.`

But I agree we can expose the attribute to override it to something else that is not used as the main message.

wsxyeah:
According to [Semantic conventions for events](https://opentelemetry.io/docs/specs/semconv/general/events/),
> Event name SHOULD be of a low-cardinality

Using `message` as event name may violate this convention, while allowing to set a dedicated event name would be decent.

mladedav:
Honestly, I don't understand the spec there as the name should uniquely identify the event and be of low cardinality at the same time.

PR #267, opened 2026-09-02
fix!: record unsigned and 128-bit integer fields as integers

mexus:
## Motivation

`u64`, `i128` and `u128` fields are not visited at all, so they fall back to
`record_debug` and reach the collector as strings. `#[instrument]` on a function
taking a `usize` or a `u64`, a `span.record("len", buf.len())`, or a plain
`info!(bytes = n)` all produce a string attribute today, while the same value as
an `i64` produces a number. Downstream that means no range queries, no
aggregations, and — in schema-inferring backends — a field whose type flips
between documents.

For `u64` this is a regression rather than a gap:

* [tokio-rs/tracing#1007] added `record_u64` to both visitors, for exactly this
  reason ("I noticed all the data gets sent to the collector as a string").
* [tokio-rs/tracing#1049], the OpenTelemetry 0.10 upgrade, removed it again with
  the note "Record `u64` values as strings as they are no longer supported in
  OpenTelemetry", because `Value::U64` had disappeared from the otel API.

That removal was the right call with the options available at the time: the
alternatives were `as i64` (wraps large values into negative numbers) or
stringify everything. Since then `opentelemetry-appender-tracing` has settled on
a third option — try the conversion, fall back to a string only when it fails —
which keeps the honest rendering for the unrepresentable case that motivated the
removal, without paying for it on the values that do fit.

[tokio-rs/tracing#1007]: https://github.com/tokio-rs/tracing/pull/1007
[tokio-rs/tracing#1049]: https://github.com/tokio-rs/tracing/pull/1049

## Solution

Implement `record_u64`, `record_i128` and `record_u128` on both `SpanEventVisitor`
and `SpanAttributeVisitor`, going through a shared helper: `i64::try_from` →
`Value::I64`, and the `Debug` rendering on failure. Not `as i64`, and not
saturating — the fallback is unreachable for every value that fits, and when it
isn't reached, it should not report a confidently wrong number. This matches
[`opentelemetry-appender-tracing`][appender], which does the same for the same
reason.

`record_bytes` is deliberately left alone: `opentelemetry::Value` has no bytes
variant, so the hex string is the honest rendering there too.

The new event methods keep both leading arms of the existing ones, so `message`
still sets the event name instead of pushing an attribute, and `log.*` fields are
still dropped as already-handled log metadata under the `tracing-log` feature —
worth noting because `log.line` arrives precisely as a `u64`.

### Breaking change

Context, since it's easy to forget: the `otel.*` special-casing lives only in
`record_str` and `record_debug`. `record_i64` doesn't have it and records a
plain attribute like any other field. So `otel.name = 5u64` renames the span
today only because there was no `record_u64` and the value fell through to
`record_debug`, while `otel.name = 5i64` has always been an attribute.

Now that `record_u64` exists, the unsigned case lands where the signed one
already did: `otel.name = <unsigned or 128-bit integer>` becomes a plain
attribute and no longer renames the span. Only callers naming a span after an
unsigned number are affected; a `&str` `otel.name` is untouched. The other three
control fields lose nothing real — `str_to_status` only accepts `"ok"`/`"error"`
and `str_to_span_kind` only the five kind names, so a numeric
`otel.status_code` or `otel.kind` already resolved to a no-op.

I followed `record_i64` rather than copying the `otel.*` arms into the new
methods, on the grounds that signedness shouldn't change what a field means —
happy to flip it if you'd rather keep the rename. It also interacts with #217,
which would put `otel.*` handling on events as well.

[appender]: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs

## Tests

Five new tests, since the attribute *type* is what nothing asserted for years:
each integer width lands as `Value::I64` on both events and span attributes
(including `span.record` after the fact, and `usize`, which routes through
`record_u64`); `u64::MAX` / `i128::MIN` / `u128::MAX` fall back to strings;
`message = 5u64` sets the event name and records no `message` attribute;
`log.*` integer fields are still dropped under `tracing-log`; and
`otel.name = 5u64` is recorded as an attribute without touching the span name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- record `u64`, `i128` and `u128` fields as OpenTelemetry integers instead of stringifying them,
+  on both events and span attributes. Since `opentelemetry::Value` has no unsigned or 128-bit
+  variant, values outside the `i64` range are still recorded as strings rather than being wrapped
+  or saturated into a wrong number.
+- [**breaking**] `otel.name = <unsigned or 128-bit integer>` no longer sets the span name and is
+  recorded as a plain attribute instead, matching how `otel.name = <i64>` has always behaved.
+
 ## [0.33.0](https://github.com/tokio-rs/tracing-opentelemetry/compare/v0.32.1...v0.33.0) - 2026-05-18
 
 ### Fixed
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -204,6 +204,21 @@ fn str_to_status(s: &str) -> otel::Status {
     }
 }
 
+/// Convert an integer of any width to an OpenTelemetry [`Value`].
+///
+/// [`Value`] has no unsigned or 128-bit variant, so integers outside the `i64` range are recorded
+/// as their `Debug` representation instead of being wrapped or saturated into a wrong number. This
+/// matches the behaviour of `opentelemetry-appender-tracing`.
+fn integer_to_value<T>(value: T) -> Value
+where
+    T: TryInto<i64> + fmt::Debug + Copy,
+{
+    match value.try_into() {
+        Ok(value) => Value::I64(value),
+        Err(_) => Value::String(format!("{value:?}").into()),
+    }
+}
+
 #[derive(Default)]
 struct SpanBuilderUpdates {
     name: Option<Cow<'static, str>>,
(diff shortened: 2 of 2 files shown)


mladedav:
I don't like the idea that the type should depend on the value.

That said there is [this document](https://opentelemetry.io/docs/specs/otel/common/attribute-type-mapping/) so it might be the correct behavior here.

I'll go through this some time later.

mexus:
Yeah, that bit feels off from a Rust point of view and I'm not thrilled about it either. But everything landing as a string is genuinely painful to work with downstream, and the spec seems to bless the value-dependent mapping, so it looks like the trade-off worth making.

No rush either way. I'm already running a fork for my own needs, so if this ends up rejected that's completely fine — I mostly opened it in case others hit the same thing.

